PR #74, opened 2019-09-10
console 0.9.0

kjeremy:
Bump console to 0.9.0 to remove dependencies on crates that depend on insta.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ serialization = []
 difference = "2.0.0"
 serde = { version = "1.0.85", features = ["derive"] }
 serde_yaml = "0.8.8"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -33,10 +33,10 @@ serialization = []
 difference = "2.0.0"
 serde = { version = "1.0.85", features = ["derive"] }
 serde_yaml = "0.8.8"
-console = "0.8.0"
+console = "0.9.0"
```
matklad: Note that we might want to use it with defaut_features = false

matklad:
friendly ping @mitsuhiko :)

kjeremy:
@mitsuhiko ?

PR #86, opened 2019-10-28
Fix typos

MichaelMcDonnell:
Fixed a couple of typos in `runtime.rs`.


max-sixty:
Thanks; mea culpa for these

PR #155, opened 2021-01-16
Inlining `difference.rs`

kper:
I could not simply use git modules because the crate uses the old `try!` deprecated macro and therefore triggers some compiler warnings, which are denied. I had to change it.

I hope it is ok, this way.


colelawrence:
Just another note here, using [`cargo deny`](https://github.com/EmbarkStudios/cargo-deny), I found that there are a few alternatives to `difference` that are maintained:

     - [`dissimilar`](https://crates.io/crates/dissimilar)
     - [`treediff`](https://crates.io/crates/treediff)
     - [`diffus`](https://crates.io/crates/diffus)

kper:
Thank you for your suggestions. It seems like the author of the crate goes with a custom solution.
Closing this PR because of #158

Issue #88, opened 2019-11-15
Insta adds trailing whitespace for blank lines in inline snapshots


matklad:
I've run `--force-update-snapshots` today. It worked flawlessly for most of the cases, expect this single problem:

![image](https://user-images.githubusercontent.com/1711539/68934475-cbe5f680-07a7-11ea-866c-b2f301e642b7.png)

Here, insta adds leading indentation to the otherwise blank line, which doesn't work great with editors which strip trailing ws on save.

PR #226, opened 2022-05-15
test: make test for #88 effective

martinvonz:
The test added in 816df8856c28 doesn't seem to test what it was
supposed to test - it still passses after replacing `l.is_empty() { 0
} else { indentation }` by just `indentation`. With the change in this
patch, the test fails with that modification.

### DIFF
--- a/insta/src/snapshot.rs
+++ b/insta/src/snapshot.rs
@@ -785,16 +785,16 @@ b"[1..];
     );
 
     let t = &"
-    a
+a
 
-    b"[1..];
+b"[1..];
     assert_eq!(
-        SnapshotContents(t.to_string()).to_inline(0),
+        SnapshotContents(t.to_string()).to_inline(4),
         "r#\"
     a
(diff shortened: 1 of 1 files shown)


martinvonz:
FYI, I ran into this while trying to fix a bug. The bug is that if you have a snapshot that looks something like`\n\nfoo\n`, then the inline snapshot will have an empty line first line since #88 (instead of line with some whitespace as it had before), but when parsing that, `insta` will skip empty lines and it will not match. I think the intent with skipping empty lines is to skip the first line after `###`. I'm going to try to fix that bug, but I haven't gotten around to it yet.

max-sixty:
@martinvonz sorry this didn't get picked up. I have now have permissions to merge things so to the extent you might still want this merged, I can do that if you can resolve conflicts...

martinvonz:
Done

max-sixty:
Thank you!

PR #296, opened 2022-10-25
feat: support postcard binary format

aminya:


src/serialization.rs
```diff
@@ -94,6 +96,12 @@ pub fn serialize_content(
             }
             rv
         }
+        #[cfg(feature = "postcard")]
+        SerializationFormat::Postcard => {
+            let data = dep_postcard::to_allocvec(&content).unwrap();
+            let data_string = format!("{:?}", data);
```
aminya, line 102: I used the debug display because the function expects a string. Otherwise, the data that is a Vec, could be directly written to the file.

PR #358, opened 2023-03-10
Simpler accept behavior on non-insta failures

max-sixty:
I wasn't sure how deliberate this choice was -- totally fine to close if it was. 

As a user, I always want it to honor the `--accept` arg.


max-sixty:
I realized this wasn't doing what it was supposed to — now fixed

PR #375, opened 2023-06-02
Pass `--color=...` to libtest

max-sixty:
Two parts to https://github.com/mitsuhiko/insta/issues/367 :
- The libtest output, which this PR solves
  - I've hand-tested it
  - Possibly this becomes moot if `cargo` is changed to handle more of the output
- The `insta` output when insta is called by libtest; e.g. the diffs. There's not much we can do from `cargo-insta`, but [`CLICOLOR_FORCE`](http://bixense.com/clicolors/) works, because `console` reads it

### DIFF
--- a/cargo-insta/src/cli.rs
+++ b/cargo-insta/src/cli.rs
@@ -892,18 +892,24 @@ fn prepare_test_runner<'snapshot_ref>(
     proc.arg("--color");
     proc.arg(color);
     proc.args(extra_args);
-    let mut dashdash = false;
+    // Items after this are passed to the test runner
+    proc.arg("--");
     if !cmd.no_quiet && matches!(test_runner, TestRunner::CargoTest) {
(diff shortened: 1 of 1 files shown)


cargo-insta/src/cli.rs
```diff
@@ -892,18 +892,24 @@ fn prepare_test_runner<'snapshot_ref>(
...
+    // Currently libtest uses a different approach to color, so we need to pass
+    // it again to get output from the test runner as well as cargo. See
+    // https://github.com/rust-lang/cargo/issues/1983 for more
+    match test_runner {
+        TestRunner::CargoTest | TestRunner::Auto => {
+            proc.arg(format!("--color={}", color));
+        }
+        _ => {}
+    };
+    dbg!(&proc.get_args());
```
max-sixty, line 912:
```suggestion
```

PR #382, opened 2023-06-12
Clippy changes to `cargo-insta`

max-sixty:
As ever for these, feel free to ignore


max-sixty:
Unfortunately the test failure is an MSRV issue (unrelated to these changes)

PR #404, opened 2023-09-16
cargo-insta: edit reject message

marcospb19:
Now the message is aware of whether there is an old snapshot or not.

### DIFF
--- a/cargo-insta/src/cli.rs
+++ b/cargo-insta/src/cli.rs
@@ -270,7 +270,7 @@ fn query_snapshot(
             println!(
                 "  {} reject     {}",
                 style("r").red().bold(),
-                style("keep the old snapshot").dim()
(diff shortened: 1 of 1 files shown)


cargo-insta/src/cli.rs
```diff
@@ -265,11 +265,21 @@ fn query_snapshot(
...
-            "  {} reject     {}",
-            style("r").red().bold(),
-            style("keep the old snapshot").dim()
-        );
+
+        if old.is_some() {
+            println!(
+                "  {} reject     {}",
+                style("r").red().bold(),
+                style("keep the old snapshot").dim()
```
max-sixty, line 273: feel free to ignore, but this would maintain the alliteration of `r`!

```suggestion
                style("retain the old snapshot").dim()
```

PR #453, opened 2024-03-03
Fix ownership issue

max-sixty:
Fixes the issue at the bottom of https://github.com/mitsuhiko/insta/pull/385#issuecomment-1975321767. 

Sorry for the mistake. 

I've added a test, which fails prior to the change.

~(FWIW I may have spotted a different bug re snapshots not updating after the first failure, I'm investigating...)~ Couldn't reproduce, don't think it's a thing.

### DIFF
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -917,3 +917,13 @@ fn test_parse_yaml_error() {
     assert!(error.contains("Failed parsing the YAML from"));
     assert!(error.contains("/bad.yaml"));
 }
+
+/// Check thot snapshots don't take ownership of the value
+#[test]
+fn test_ownership() {
+    // Range is non-copy
+    use std::ops::Range;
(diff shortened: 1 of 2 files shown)


src/snapshot.rs
```diff
@@ -917,3 +917,13 @@ fn test_parse_yaml_error() {
     assert!(error.contains("Failed parsing the YAML from"));
     assert!(error.contains("/bad.yaml"));
 }
+
+/// Check thot snapshots don't take ownership of the value
```
kennethloeffler, line 921:
```suggestion
/// Check that snapshots don't take ownership of the value
```

kennethloeffler (commented):
I've swapped this branch into our project and can confirm that it does fix the issue. I did notice an unfortunate typo in a comment, but otherwise LGTM

PR #455, opened 2024-03-05
Allow `debug` expressions in redaction macros

max-sixty:
It can be a bit confusing when some macros have features which others don't. So this adds debug expressions to the redaction macros.

(Though the calls do get quite long; possibly we want to recommend something like `description` in place of `debug` throughout?)


max-sixty:
This is rebased and ready for merging / review

max-sixty:
I forgot about this and did it in https://github.com/mitsuhiko/insta/pull/756! so closing

PR #472, opened 2024-04-03
Allow trailing commas in inline snapshots

narpfel:
After using https://github.com/mitsuhiko/insta-cmd/pull/12 for a bit, I noticed that `cargo insta test` would not pick up on mismatches in inline snapshots with trailing commas.

This PR fixes that and also allows trailing commas for inline snapshots in other macros.

This PR uses code from #454 by @max-sixty.

### DIFF
--- a/insta/src/macros.rs
+++ b/insta/src/macros.rs
@@ -219,7 +219,7 @@ macro_rules! assert_compact_json_snapshot {
     };
 }
 
-// The macro handles optional trailing commas for file snapshots.
+// This macro is expected to handle optional trailing commas.
 #[doc(hidden)]
 #[macro_export]
 macro_rules! _assert_serialized_snapshot {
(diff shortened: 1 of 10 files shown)


insta/src/macros.rs
```diff
@@ -321,7 +319,7 @@ macro_rules! assert_debug_snapshot {
 // the value. This allows us to implement other macros with a small wrapper. All
 // snapshot macros eventually call this macro.
 //
-// The macro handles optional trailing commas in file snapshots.
+// This macro is expected to handle trailing commas.
```
max-sixty, line 322:
```suggestion
// This macro handles trailing commas.
```

max-sixty (commented):
Awesome @narpfel

Is the comma be removed when a snapshot is written? If so, do we want that or do we prefer to retain it?

We don't have great tests for this sort of thing IIRC, so it's worthwhile checking by hand that the replacement behavior is what's expected (or ofc +1 to adding a test...)

(@mitsuhiko is the maintainer so this is just a comment from another contributor)

narpfel:
> Is the comma be removed when a snapshot is written? If so, do we want that or do we prefer to retain it?

The comma is kept, for example in this test: https://github.com/mitsuhiko/insta/pull/472/commits/6b28cb349736ccdbe5ea2417d978fb535fb9a882#diff-42101e5a83cd4bee69fc68a5258dfbeafa0c657945e327f0aee0fd05ab309d58R35-R38

insta/src/macros.rs
narpfel: Like this? db010fb8439d8dc43e31b644cb9db96a9e958678

max-sixty: Yes cool! (FYI GH also allows "Accept suggestion")

max-sixty:
> The comma is kept, for example in this test: [6b28cb3#diff-42101e5a83cd4bee69fc68a5258dfbeafa0c657945e327f0aee0fd05ab309d58R35-R38](https://github.com/mitsuhiko/insta/commit/6b28cb349736ccdbe5ea2417d978fb535fb9a882#diff-42101e5a83cd4bee69fc68a5258dfbeafa0c657945e327f0aee0fd05ab309d58R35-R38)

Perfect!

PR #479, opened 2024-05-02
Improve docs for snapshot updates

max-sixty:
This attempts to clarify when snapshots are written to pending `.snap.new` files vs. to overwriting `.snap` files vs. doing nothing to snapshots.

It also changes `.with_extension(...)`, lmk if you prefer those on a separate PR

### DIFF
--- a/cargo-insta/src/walk.rs
+++ b/cargo-insta/src/walk.rs
@@ -35,8 +35,7 @@ pub(crate) fn find_snapshots<'a>(
             let fname = e.file_name().to_string_lossy();
             if fname.ends_with(".new") {
                 let new_path = e.into_path();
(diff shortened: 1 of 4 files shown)


max-sixty:
Requires #481 to pass

max-sixty:
Starting again, I would change some of the names
- `new` could be `review` or `pending`, currently doesn't correspond to the other uses of this in `cargo-insta`, and it sounds like what `unseen` does
-  `always` could be `accept`
- Is `unseen` useful?!

...but it would be an annoying backward-compat cycle given we want to support different versions of `cargo-insta` & the config files.

max-sixty:
This is ready to merge

PR #480, opened 2024-05-02
Test `cargo doc` in CI

max-sixty:
Assert no warnings (will fail at first)


max-sixty:
This is ready to merge & not controversial

Issue #477, opened 2024-04-29
require_full_match mode always fails with matching snapshots


dspicher:
### What happened?

When I run `INSTA_REQUIRE_FULL_MATCH=1 cargo test` on a [dummy snapshot test case](https://github.com/dspicher/insta-require-full-match), it fails with the following output:

<details>
  <summary>Output</summary>
  
```shell
$ INSTA_REQUIRE_FULL_MATCH=1 cargo t
   Compiling insta-tests v0.1.0 (/home/dominik/purgable/insta-tests)
    Finished test [unoptimized + debuginfo] target(s) in 0.44s
     Running unittests src/lib.rs (target/debug/deps/insta_tests-3c106da6b8f1533f)

running 1 test
test tests::test_insta ... FAILED

failures:

---- tests::test_insta stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: src/snapshots/insta_tests__tests__insta.snap
Snapshot: insta
Source: src/lib.rs:5
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: 1 + 1
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
          - │ snapshots are matching
────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
thread 'tests::test_insta' panicked at .../.cargo/registry/src/index.crates.io-6f17d22bba15001f/insta-1.38.0/src/runtime.rs:563:9:
snapshot assertion for 'insta' failed in line 5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    tests::test_insta

[...]

PR #483, opened 2024-05-02
Fix snapshot & module name extraction

### LINKED ISSUE #401: cargo-insta review emits "Snapshot" with everything after a period removed
### What happened?

In `cargo-insta review` I see

```
Reviewing [1/1] foo@0.1.0:
Snapshot file: foo/src/snapshots/go1.20.5.snap
Snapshot: go1
Source: foo/src/lib.rs:290
```

It looks as though the intent is to strip off ".snap" but instead everything after the first period is stripped.

### Reproduction steps

_No response_

### Insta Version

1.31.0

### rustc Version

1.72.0

### What did you expect?

The full snapshot name should be shown.

max-sixty:
Fixes #401 as well as most (but not all) of #477

### DIFF
--- a/insta/src/snapshot.rs
+++ b/insta/src/snapshot.rs
@@ -323,31 +323,11 @@ impl Snapshot {
             buf.push_str(&line);
         }
 
-        let module_name = p
-            .file_name()
-            .unwrap()
-            .to_str()
-            .unwrap_or("")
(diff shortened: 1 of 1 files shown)


max-sixty:
This is ready to merge & I don't think is controversial

PR #484, opened 2024-05-02
Be robust to removing pending snapshots during review

### LINKED ISSUE #380: Being more permissive to files changing under insta

max-sixty:
Closes #380

### DIFF
--- a/cargo-insta/src/container.rs
+++ b/cargo-insta/src/container.rs
@@ -164,6 +164,18 @@ impl SnapshotContainer {
     }
 
     pub(crate) fn commit(&mut self) -> Result<(), Box<dyn Error>> {
+        // Try removing the snapshot file. If it fails, it's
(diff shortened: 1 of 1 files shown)


max-sixty:
Ready to merge, fairly non-controversial I think

PR #502, opened 2024-06-14
Satisfy latest clippy

max-sixty:


max-sixty:
@mitsuhiko we can merge this one and then have `main` passing...

PR #517, opened 2024-07-01
Upgrade to syn 2

max-sixty:
Need to upgrade `structopt` and then we'll reduce the compile time by coalescing on a single `syn`

~Edit: possibly we need to wait for a newer MSRV~

Ready to merge

### DIFF
--- a/cargo-insta/Cargo.toml
+++ b/cargo-insta/Cargo.toml
@@ -20,7 +20,7 @@ structopt = { version = "0.3.26", default-features = false }
 serde = { version = "1.0.117", features = ["derive"] }
 serde_json = "1.0.59"
 proc-macro2 = { version = "1.0.24", features = ["span-locations"] }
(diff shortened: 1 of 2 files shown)


cargo-insta/Cargo.toml
```diff
@@ -19,8 +19,8 @@ console = "0.15.4"
 structopt = { version = "0.3.26", default-features = false }
 serde = { version = "1.0.117", features = ["derive"] }
 serde_json = "1.0.59"
-proc-macro2 = { version = "1.0.24", features = ["span-locations"] }
-syn = { version = "1.0.50", features = ["full", "visit", "extra-traits"] }
+proc-macro2 = { version = "1.0.60", features = ["span-locations"] }
```
max-sixty, line 22: Required because of https://github.com/rust-lang/rust/issues/113152

PR #523, opened 2024-07-07
Bump `pest` versions

max-sixty:
Because older versions of pest can support really old versions of `proc-macro2`, this was causing some errors in min-versions checks (for example https://github.com/mitsuhiko/insta/actions/runs/9829003042/job/27133598634?pr=518#step:6:229).

It doesn't solve any 'actual' problems, it just fixes the checks, but it's also very low cost. Pest 2.4.0 was out two years ago


max-sixty:
The source of the error was actually different, will close

PR #524, opened 2024-07-07
Add `cargo.lock`, use `cargo-msrv` to verify MSRV

max-sixty:
This commit does a few things:
- Adds `Cargo.lock` to the repo. Maybe others have different opinions; though it is [now recommended](https://github.com/rust-lang/cargo/commit/3bb02f23a23f90261b44a107e1b8821a504bac08#diff-3388c611c3bcae8ef4efe0ec5c9f5e22682b0db86c9e7137e70e5ba7d5e456c0R11), and it solves a bunch of the issues around `Cargo.lock.msrv`, closing #498. Without this, the remainder of this PR isn't feasible in its current state.
- Uses `cargo-msrv` to verify the MSRV specs in the `Cargo.toml` files in CI, simplifying CI jobs
- Adds 1.61.0 as the MSRV for `cargo-insta`
- Bumps the MSRV for `insta` to 1.57.0, from 1.51.0, allowing us to remove some of the hacks around named profiles. 1.57.0 was released Dec 2021.

PR #525, opened 2024-07-07
Check whether building without a cache passes

max-sixty:
Currently when I run `cargo update && cargo build` locally, I get failures around `similar 2.5.0`. I'm not sure why CI has been passing; this is a test to check whether it's the cache which is holding an existing version of dependencies.

#524 adds the lockfile to the repo, which would reduce the change of a local dev break

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,8 +22,13 @@ jobs:
         run: make test
 
   test-temp-no-cache:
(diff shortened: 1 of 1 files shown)


max-sixty:
This ended up being a corrupt local registry, but out of it comes https://github.com/mitsuhiko/insta/pull/526 to fix a test on Windows

PR #536, opened 2024-07-28
Simplify cargo metadata collection

max-sixty:
It was previously a bit disparate and difficult to verify. This simplifies & unifies the code.

It builds on https://github.com/mitsuhiko/insta/pull/532, so that should merge first.

I'm still not 100% confident — ideally we'd have some integration tests with various workspace configurations...


max-sixty:
I realize I forgot about this and re-implemented half of it in #532, apologies. Given it makes the code simpler and easier to review, I'll copy the improvements over to there and close this.

PR #537, opened 2024-07-28
New integration test approach

max-sixty:
~Very draft — but shows a reasonably scalable way to add integration tests. There are some TODOs at the top.~

~This would help with issues such as #531~

Replaced the existing integration tests with a new approach, which:
- Copies a set of files defined inline into a temp dir, executes tests, and then shows the diff
- Allows adding integration tests much more easily — including whole repo tests — so problems we've had with `cargo-insta` are very tractable to solve
- Adds the repro of @regexident as a test for #531 (which currently will fail; requires merging this first and then we'll un-skip the test)
- ...while it might be overkill for the existing tests, it already allows us to remove some complicated code which was handling tests-running-tests. And it lets us do much more.

~This is stacked on #540, that should merge first~ merged


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to insta and cargo-insta are documented here.
 
 ## 1.40.0
 
+- Inline snapshots now use with the required number of `#` to escape the
```
regexident, line 7: The beginning of this sentence has a weird grammar.

max-sixty: Thanks, fixed

PR #549, opened 2024-08-02
Improve integration tests

max-sixty:
This splits out the changes in the test code from #532 in order to make that diff smaller.

It divides  the building of the test case files from the test case files on disk + running the test, which make the structs simpler & more obvious


max-sixty:
Going to merge this now to make the diff of #532 smaller (lmk if this is over-eager)

PR #551, opened 2024-08-03
Add missing changlog entries for 1.40

max-sixty:


tisonkun:
@max-sixty Thanks for catching up these. Is there an estimation for 1.40.0 release?

max-sixty:
That's up to @mitsuhiko

PR #553, opened 2024-08-04
Update snapshot file

a-kenji:
This updates a test snapshot.


max-sixty:
What's the goal of this?

a-kenji:
Ah, apologies. It was an attempt to run the test suite, since the `Makefile` was giving unhelpful errors.
I managed to run it now and see this is intentional.

When running just `cargo test` the tests fail, which is surprising at first.

max-sixty:
What's the error you get when running `make test`? It passes on my end and in CI, but we should ensure it passes on all systems, so that's helpful feedback if it fails...

PR #563, opened 2024-08-11
Enforce starting newlines

max-sixty:
This is an example of what I've been suggesting at https://github.com/mitsuhiko/insta/pull/506 (+ some linked issues):
- ~Stacked on https://github.com/mitsuhiko/insta/pull/528 so that would need to merge first~
- Enforces the correct number of newlines at the start of snapshots. The start is easier because we don't need to worry about editor issues in trailing newlines (and don't need to worry about how `Lines` deals ambiguously with trailing new lines)
- Works for both inline & file snapshots
- Adds a `matches_legacy` method on snapshot contents — this encapsulates the older formats we still support and warns about them

~It's not ready to merge yet, because we don't seem to actually support `--force-update-snapshots` on inline values, which is somewhat required. But I'm publishing for feedback & to put some weight on #528 in case we don't want to pursue.~


insta/src/serialization.rs
```diff
@@ -166,15 +150,23 @@ fn test_yaml_serialization() {
...
-    ---
     env:
       - ENVIRONMENT
       - production
     cmdline:
       - my-tool
       - run
     "###);
+
+    // Old approach with leading `---`:
```
max-sixty, line 163: TODO: move to an integration test, given this was correct in #528 but then overwritten in this PR's `--force-update-snapshots` run

max-sixty: Done

max-sixty:
One downside (but balanced against lots of upsides!) is that because we insert a newline at the start of a multiline string, a literal string can look a bit odd:

```rust
assert_snapshot!("
foo
bar
", @r"

foo
bar
");
```

Notice that because the original literal has a newline after its `"`, and we always add a newline to the original value, there are then two newlines after the opening quote of the result.

This is mostly only noticeable in examples: very few generated values start with a newline!

max-sixty:
I'm tempted to merge this as there's no downside to enforce starting newlines; and it simplifies the code overall

lmk any objections...

max-sixty:
merging but one click to revert if anyone disagrees

PR #564, opened 2024-08-11
Collapse empty methods

max-sixty:
Was there a reason to have all of these rather than just using the fields?


max-sixty:
Ok! I don't feel strongly, I'll close.

Though out of interest, if we want to add checks then surely we can change then later? These are only an internal interface...

PR #573, opened 2024-08-31
draft: Prevent lots of empty inline snapshots with `--force-update-snapshots` without `--accept`

max-sixty:
I realize what you mean in the prior comment. Currently running with `--accept` is kinda painful. I'm not sure how common that is, but would be nice to fix.

Had a go for 30 mins, need to run, but pushing an initial effort and a pointer to the thing that needs fixing

### DIFF
--- a/insta/src/runtime.rs
+++ b/insta/src/runtime.rs
@@ -667,8 +667,23 @@ pub fn assert_snapshot(
     if pass {
         ctx.cleanup_passing()?;
 
-        if tool_config.force_update_snapshots() {
-            ctx.update_snapshot(new_snapshot)?;
+        if dbg!(tool_config.force_update_snapshots()) {
+            // Inline snapshots currently create lots of files even when they
+            // match. So we don't update them if they fully match.
+            //
+            // Note that there's a check down the stack on whether the files
+            // match for file snapshots; probably we should combine that check
+            // with `matches_fully` and then use a single check for whether we
+            // force update snapshots. Currently the bad user experience is for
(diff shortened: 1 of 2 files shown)


max-sixty:
OK, I think I understand the underlying problem here. In a test like:

```
assert_snapshot!("test", @r#test#");
```

- The macro invocation only sees the `test` value; it doesn't know about the `r#...#`
- `insta` can't find out about the `r#...#`, only `cargo-insta`, because only `cargo-insta` has the rust parser
- We need to compare the full `r#...#` to correctly decide whether a snapshot requires updating when `--force-update-snapshots` is passed, and we need to do that with `insta` only
- Unless we can run `matches_fully` correctly, we default to creating pending snapshots for snapshots which don't need updating, which is what the existing `master` does

A couple of ideas, none are great:
- Have `--force-update-snapshots` accept by default. This is reasonable UX — it's generally something folks are not running on every loop, but rather an occasional clean-up. 
  - Currently, running `--force-update-snapshots --accept` is a reasonable experience — it prints a few too many names of snapshots it's updating, but that's the only cost. 
  - It also sounds like that's what it's doing already, or we could pick a new name `--force-accept` if we wanted
- Revert https://github.com/mitsuhiko/insta/pull/569, give up on force-updating inline snapshots.
- Offer some additional option; `--force-update-inline-snapshots`. This would be a shame.

My current thinking is to revert #569, and then default to `--force-update-snapshots` accepting over a couple of versions.

max-sixty:
I think this is broadly fixed in #581

PR #581, opened 2024-09-01
Adjust force-update on inline snapshots to only view within the string

max-sixty:
This solves the issue in #573 for the moment:
- When `--force-update-snapshots` in passed, we only update inline snapshots when there's some difference _within_ the string, such as an additional linebreak at the start or the end
- We no longer update the surrounding hashes. In the existing code, this happened because we were writing too many inline snapshots, not because we were actually checking the number of hashes
  - Checking the number of hashes isn't easy to do — reason outlined at https://github.com/mitsuhiko/insta/pull/573#issuecomment-2323500808
- The main change in the code is that we store the unnormalized snapshot and then normalize when we need to. The existing code stored the normalized version, which meant we couldn't evaluate whether the unnormalized versions were different

It's a decent number of changes, but will integrate nicely with https://github.com/mitsuhiko/insta/pull/563.

~(FYI we currently don't look at the indentation, but we could adjust this)~ Now indentation works too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to insta and cargo-insta are documented here.
 
 ## 1.41.0
 
+- `--force-update-snapshots` has more conservative and consistent behavior for
+  inline snapshots. As a side-effect of this, only the content within the inline
+  snapshot delimiters are assessed for changes, not the delimiters (e.g. `###`).
+  #581
+
 ## 1.40.0
 
 - `cargo-insta` no longer panics when running `cargo insta test --accept --workspace`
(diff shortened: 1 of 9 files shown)


max-sixty:
OK, I think this is in a pretty good state now — again, more changes than ideal — but I think moving the library to be simpler, fewer corners

max-sixty:
@mitsuhiko this one is top of my list for you to review when you get the chance to review things; think it's a much better experience for users; somewhat to your point when merging the PR that enabled forcing inline snapshot updates

max-sixty:
@mitsuhiko I think the benefits of this are large enough, and I would like to minimize conflicts with #489, that it's worth me merging without waiting longer for a review. The main drawback is potentially one of performance — the current code does many more allocations, albeit in non-perf-sensitive code.

I recognize you gave me merge permissions with the understanding I would be careful, so:
- For the future, was merging this OK or would you have preferred I wait?
- I'll address any objections
- If you strongly object, I'll own reverting and rectifying any conflicts

PR #584, opened 2024-09-02
Add some docstrings & a couple of renamings

max-sixty:


max-sixty:
Merging, lmk any objections post

PR #609, opened 2024-09-22
Replace complicated name clash test with simple integration test

max-sixty:


max-sixty:
The existing test prevented insta tests themselves from passing on `--force-update-snapshots`, so merging

PR #614, opened 2024-09-25
Consolidate onto `INSTA_WORKSPACE_ROOT`

### LINKED ISSUE #575: `INSTA_WORKSPACE_ROOT` should also work with `assert_snapshot!` at runtime, for manual test binaries
I am evaluating insta with a project that uses [Buck2](https://github.com/facebook/buck2) (and also a very happy user using it with Cargo as part of [Jujutsu](https://github.com/martinvonz/jj)). When using a project with Buck2, Cargo is not used, and so tests for example are run "manually" by the build system itself after compilation with `--cfg=test`

Currently, I'm using inline snapshots, with a test like this:

```rust
#[cfg(test)]
mod tests {
    fn split_words(s: &str) -> Vec<&str> {
        s.split_whitespace().collect()
    }

    #[test]
    fn test_split_words() {
        let words = split_words("goodbye from the other side");
        insta::assert_snapshot!(format!("{:?}", words), @"");
    }
}
```

By default, when I compile this with `buck2 build`, I get an error like so:

```
error: environment variable `CARGO_MANIFEST_DIR` not defined at compile time
   --> src/qq-cli/commands/script.rs:148:9
    |
148 |         insta::assert_snapshot!(format!("{:?}", words), @"");
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Buck2 executes all tests from the root of the project, so luckily I can set `CARGO_MANIFEST_DIR=.` during the compilation of these tests which solves the problem. But it's a bit ugly I think.

While using Code Search to browse the source code to figure out how the macro worked, it looks like `INSTA_WORKSPACE_ROOT` also has a similar function. It would at least be more *explanatory* when I write `INSTA_WORKSPACE_ROOT=.` in the source code why it exists. However, it seems like it only works when you use `cargo insta`.

Alternatively, some other more general form or utility for `assert_snapshot!()` to work in the presence of non-Cargo systems would be nice.

max-sixty:
Intended to fix #575. Though after some iterations, it ends up being a code-clean-up. I'm not sure how `CARGO_MANIFEST_DIR` is leaking through — it's only referenced by `env!` rather than `env::var`, so should only matter at compile time.

Added a pretty thorough integration test (albeit maybe a bit too complicated)

### DIFF
--- a/cargo-insta/tests/main.rs
+++ b/cargo-insta/tests/main.rs
@@ -1215,7 +1215,6 @@ fn test_insta_workspace_root() {
     #[cfg(test)]
     mod tests {
         use insta::assert_snapshot;
-        use super::*;
 
         #[test]
         fn test_snapshot() {
--- a/cargo-insta/tests/main.rs
+++ b/cargo-insta/tests/main.rs
@@ -1,3 +1,6 @@
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::Command;
 /// Integration tests which allow creating a full repo, running `cargo-insta`
 /// and then checking the output.
 ///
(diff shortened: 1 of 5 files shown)


max-sixty:
I'll merge now since it's just a refactor and there's a fairly complete test.

This closes #675 insofar as I can't see how that issue exists in the proposed code. But I also can't see how it existed in the recent code, so I'm not confident it's fixed. If there's a reproducible failure, happy to get it working.

PR #618, opened 2024-09-26
docs: improved the linking inside and outside of the crate

CommanderStorm:
While browsing the docs on docs.rs, I noticed that there are a few items which are not linked.
This PR tries to improve this situation.

Drawing attention to one thing which might be controversial:
In the current codebase, the usage of `[´foo´](Self::foo)` and `[´Self::foo´]` (and related links) are a bit inconsistent. I have changed all to the `[´Self::foo´]` format. Please let me know if you prefer the first and I can change them to that format.

I have verified this change via adding the following (somewhat agressive) docs-related lints and running `RUSTDOCFLAGS="--cfg=docsrs" cargo +nightly doc --no-deps --all-features` and `cargo clippy --all-features --all`.

```rust
#![warn(clippy::doc_markdown)]
#![warn(rustdoc::all)]
```

### DIFF
--- a/insta/src/env.rs
+++ b/insta/src/env.rs
@@ -328,7 +328,7 @@ impl ToolConfig {
         self.snapshot_update
     }
 
-    /// Returns the value of [`Self::glob_fail_fast`]
+    /// Returns whether the glob should fail fast, as snapshot failures within the glob macro will appear only at the end of execution unless `glob_fail_fast` is set.
     #[cfg(feature = "glob")]
     pub fn glob_fail_fast(&self) -> bool {
         self.glob_fail_fast
--- a/insta/src/env.rs
+++ b/insta/src/env.rs
@@ -328,7 +328,7 @@ impl ToolConfig {
         self.snapshot_update
     }
 
-    /// Returns the value of glob_fail_fast
+    /// Returns the value of [`Self::glob_fail_fast`]
     #[cfg(feature = "glob")]
     pub fn glob_fail_fast(&self) -> bool {
         self.glob_fail_fast
--- a/insta/src/lib.rs
+++ b/insta/src/lib.rs
@@ -1,3 +1,6 @@
+#![warn(clippy::doc_markdown)]
+#![warn(rustdoc::all)]
+
 //! <div align="center">
 //!  <img src="https://github.com/mitsuhiko/insta/blob/master/assets/logo.png?raw=true" width="250" height="250">
 //!  <p><strong>insta: a snapshot testing library for Rust</strong></p>
--- a/insta/src/lib.rs
+++ b/insta/src/lib.rs
@@ -80,7 +80,7 @@
 //!   [`std::fmt::Display`] outputs, often strings.
 //! - [`assert_debug_snapshot!`] for comparing [`Debug`] outputs of values.
 //!
-//! The following macros require the use of [`serde`]'s [`serde::Serialize`]:
+//! The following macros require the use of [`serde::Serialize`]:
 //!
 #![cfg_attr(
     feature = "csv",
--- a/insta/src/lib.rs
+++ b/insta/src/lib.rs
@@ -100,11 +100,11 @@
 )]
 #![cfg_attr(
     feature = "json",
-    doc = "- [`assert_json_snapshot!`] for comparing `JSON` serialized output. (requires the `json` feature)"
+    doc = "- [`assert_json_snapshot!`] for comparing JSON serialized output. (requires the `json` feature)"
 )]
 #![cfg_attr(
     feature = "json",
-    doc = "- [`assert_compact_json_snapshot!`] for comparing `JSON` serialized output while preferring single-line formatting. (requires the `json` feature)"
+    doc = "- [`assert_compact_json_snapshot!`] for comparing JSON serialized output while preferring single-line formatting. (requires the `json` feature)"
 )]
 //!
 //! For macros that work with [`serde`] this crate also permits redacting of
@@ -171,7 +171,7 @@
 //! The following features exist:
 //!
 //! * `csv`: enables CSV support (via [`serde`])
-//! * `json`: enables `JSON` support (via [`serde`])
+//! * `json`: enables JSON support (via [`serde`])
 //! * `ron`: enables RON support (via [`serde`])
 //! * `toml`: enables TOML support (via [`serde`])
 //! * `yaml`: enables YAML support (via [`serde`])
--- a/insta/src/lib.rs
+++ b/insta/src/lib.rs
@@ -177,7 +177,7 @@
 //! * `yaml`: enables YAML support (via [`serde`])
 //! * `redactions`: enables support for redactions
 //! * `filters`: enables support for filters
-//! * `glob`: enables support for globbing ([`crate::glob!`])
+//! * `glob`: enables support for globbing ([`glob!`])
 //! * `colors`: enables color output (enabled by default)
 //!
 //! For legacy reasons the `json` and `yaml` features are enabled by default in
--- a/insta/src/lib.rs
+++ b/insta/src/lib.rs
@@ -77,7 +77,7 @@
 //! This crate exports multiple macros for snapshot testing:
 //!
 //! - [`assert_snapshot!`] for comparing basic snapshots of
-//!   [`std::fmt::Display`] outputs, often strings.
+//!   [`Display`](std::fmt::Display) outputs, often strings.
 //! - [`assert_debug_snapshot!`] for comparing [`Debug`] outputs of values.
 //!
 //! The following macros require the use of [`serde::Serialize`]:
--- a/insta/src/lib.rs
+++ b/insta/src/lib.rs
@@ -7,7 +7,7 @@
 //!
 //! Snapshots tests (also sometimes called approval tests) are tests that
 //! assert values against a reference value (the snapshot).  This is similar
-//! to how `assert_eq!` lets you compare a value against a reference value but
+//! to how [`assert_eq!`] lets you compare a value against a reference value but
 //! unlike simple string assertions, snapshot tests let you test against complex
 //! values and come with comprehensive tools to review changes.
 //!
@@ -77,10 +77,10 @@
 //! This crate exports multiple macros for snapshot testing:
 //!
 //! - [`assert_snapshot!`] for comparing basic snapshots of
-//!   [`Display`](std::fmt::Display) outputs, often strings.
+//!   [`std::fmt::Display`] outputs, often strings.
 //! - [`assert_debug_snapshot!`] for comparing [`Debug`] outputs of values.
 //!
-//! The following macros require the use of serde's [`Serialize`](serde::Serialize):
+//! The following macros require the use of [`serde`]'s [`serde::Serialize`]:
 //!
 #![cfg_attr(
     feature = "csv",
@@ -100,11 +100,11 @@
 )]
 #![cfg_attr(
     feature = "json",
-    doc = "- [`assert_json_snapshot!`] for comparing JSON serialized output. (requires the `json` feature)"
+    doc = "- [`assert_json_snapshot!`] for comparing `JSON` serialized output. (requires the `json` feature)"
 )]
 #![cfg_attr(
     feature = "json",
-    doc = "- [`assert_compact_json_snapshot!`] for comparing JSON serialized output while preferring single-line formatting. (requires the `json` feature)"
+    doc = "- [`assert_compact_json_snapshot!`] for comparing `JSON` serialized output while preferring single-line formatting. (requires the `json` feature)"
 )]
 //!
 //! For macros that work with [`serde`] this crate also permits redacting of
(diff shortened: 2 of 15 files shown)


insta/src/env.rs
```diff
@@ -328,7 +328,7 @@ impl ToolConfig {
         self.snapshot_update
     }
 
-    /// Returns the value of glob_fail_fast
+    /// Returns the value of [`Self::glob_fail_fast`]
```
max-sixty, line 331: I think this is very tautological (all these methods' docstrings here are somewhat tautological now I'm looking at them). But at least without the quote it refers to the setting rather than the method itself; so we should revert this line

insta/src/lib.rs
```diff
@@ -77,10 +77,10 @@
 //! This crate exports multiple macros for snapshot testing:
 //!
 //! - [`assert_snapshot!`] for comparing basic snapshots of
-//!   [`Display`](std::fmt::Display) outputs, often strings.
+//!   [`std::fmt::Display`] outputs, often strings.
```
max-sixty, line 80: How does this appear in markdown? This is more user-focused docs, so I think we just want the `Display` to appear, as though we're speaking it out loud.

(most of the other changes are fine, that's more API docs)

insta/src/lib.rs
```diff
@@ -77,10 +77,10 @@
...
 //! - [`assert_debug_snapshot!`] for comparing [`Debug`] outputs of values.
 //!
-//! The following macros require the use of serde's [`Serialize`](serde::Serialize):
+//! The following macros require the use of [`serde`]'s [`serde::Serialize`]:
```
max-sixty, line 83: Same here; though could do

```suggestion
//! The following macros require the use of [`serde::Serialize`]:
```

insta/src/lib.rs
```diff
@@ -100,11 +100,11 @@
 )]
 #![cfg_attr(
     feature = "json",
-    doc = "- [`assert_json_snapshot!`] for comparing JSON serialized output. (requires the `json` feature)"
+    doc = "- [`assert_json_snapshot!`] for comparing `JSON` serialized output. (requires the `json` feature)"
```
max-sixty, line 103: JSON is common language, there's no need to quote it

insta/src/lib.rs
```diff
@@ -170,27 +170,28 @@
...
-//! * `yaml`: enables YAML support (via serde)
+//! * `csv`: enables CSV support (via [`serde`])
+//! * `json`: enables `JSON` support (via [`serde`])
+//! * `ron`: enables RON support (via [`serde`])
+//! * `toml`: enables TOML support (via [`serde`])
+//! * `yaml`: enables YAML support (via [`serde`])
 //! * `redactions`: enables support for redactions
 //! * `filters`: enables support for filters
-//! * `glob`: enables support for globbing ([`glob!`])
+//! * `glob`: enables support for globbing ([`crate::glob!`])
```
max-sixty, line 180: Did the existing version work OK? If so, the proposed version seems over-specified

insta/src/lib.rs
```diff
@@ -258,8 +259,10 @@
 //! opt-level = 3
 //! ```
 //!
-//! You can also disable the default features of `insta` which will cut down on
+//! You can also disable the default features of [`insta`] which will cut down on
 //! the compile time a bit by removing some quality of life features.
+//!
+//!  [`insta`]: https://docs.rs/insta
```
max-sixty, line 265: This is self-referential, no? I also think this is over-structured...

insta/src/macros.rs
```diff
@@ -74,14 +74,14 @@ macro_rules! assert_toml_snapshot {
...
+/// Asserts a [`serde::Serialize`] snapshot in YAML format.
 ///
 /// **Feature:** `yaml`
 ///
-/// The value needs to implement the `serde::Serialize` trait and the snapshot
+/// The value needs to implement the [`serde::Serialize`] trait and the snapshot
 /// will be serialized in YAML format.  This does mean that unlike the debug
 /// snapshot variant the type of the value does not appear in the output.
-/// You can however use the `assert_ron_snapshot!` macro to dump out
+/// You can however use the [`crate::assert_ron_snapshot!`] macro to dump out
```
max-sixty, line 84: All these `crate` specs are too much structure iff the links work without them. Fine if they're required on balance

max-sixty (commented):
Overall great, but I think we should scale it back about 30% — anything that doesn't require `crate::` or `self::` or `std::fmt`, etc could be lighter. Is that reasonable?

Happy to try enabling some of the lints if they are manageable to work with... We can ignore the `vendored` path though; that code won't see much development.

insta/src/env.rs
CommanderStorm: Okay, what do you think about this doccomment instead?

```suggestion
    /// Returns if the glop should fail fast
    /// 
    /// Snapshot failures within the glob macro appear (unless `glob_fail_fast` is set) at the end of of the execution.
```

insta/src/lib.rs
CommanderStorm: I think this appears as documented. At least for `crate::`, it does.
Lets revert this

```suggestion
//!   [`Display`](std::fmt::Display) outputs, often strings.
```

insta/src/lib.rs
CommanderStorm: Yes and no.

We reference `similar` here => linking `insta` too looks nicer..
image

insta/src/macros.rs
CommanderStorm: Jea.. I have no clue why they are required..
(the other references are equivalently nessesary sadly.. maybe marcros and scoping? Would need to debug further, but currently don't have time for diving into rustdoc^^)

This seems more like a bug, right?

image

insta/src/lib.rs
CommanderStorm: Interestingly, what does not work in https://github.com/mitsuhiko/insta/pull/618/files#r1777903673 does work here.. 🤯

I don't understand rustdoc's scoping today..

CommanderStorm:
> Drawing attention to one thing which might be controversial:
In the current codebase, the usage of `[´foo´](Self::foo)` and `[´Self::foo´]` (and related links) are a bit inconsistent. I have changed all to the `[´Self::foo´]` format. Please let me know if you prefer the first and I can change them to that format.

> I think we should scale it back about 30% — anything that doesn't require `crate::` or `self::` or `std::fmt`, etc could be lighter. Is that reasonable?

Just that we are on the same page.
I should handle this via the first option (`[´foo´](Self::foo)`), right?

I have hidden the `std::fmt` and `crate::` calls as such.

For `Self::` though this gives a bit of additional info (where is this located in relateion to what I am reading) which I whink might be helpfull.

CommanderStorm:
> We can ignore the `vendored` path though

Read: Please remove the changes from said code?

insta/src/env.rs
max-sixty: Yes, if we can make that into sentence that would be great; the new line you added has the useful info!

max-sixty:
> Read: Please remove the changes from said code?

No need to revert! But no need to do any more on it / have strict lints etc.

max-sixty:
> Just that we are on the same page.
> I should handle this via the first option ([´foo´](Self::foo)), right?
> 
> I have hidden the std::fmt and crate:: calls as such.
> 
> For Self:: though this gives a bit of additional info (where is this located in relateion to what I am reading) which I whink might be helpfull.

Yes! I think the current code is a great balance, thank you!

insta/src/env.rs
CommanderStorm: Making this into one sentence sounds fine. What do you think of the following?

```suggestion
    /// Returns whether the glob should fail fast, as snapshot failures within the glob macro will appear only at the end of execution unless `glob_fail_fast` is set.
```

insta/src/macros.rs
max-sixty: Is the feature enabled?

CommanderStorm: Yes, have just double checked myself. I did activate all features:

```bash
cargo doc --all-features --all --keep-going 
```

CommanderStorm: This is also reproducible via `make doc`, aka
```bash
RUSTC_BOOTSTRAP=1 RUSTDOCFLAGS="--cfg=docsrs" cargo doc --no-deps --all-features
```

max-sixty:
I can't repro the error above re `assert_ron_snapshot`, but I do get this error:

```
cargo clippy --all-features --all-targets
warning: backticks are unbalanced
   --> insta/src/content/yaml/vendored/emitter.rs:253:5
    |
253 |   /// Strings starting with any of the following characters must be quoted.
    |  _____^
254 | | /// `:`, `&`, `*`, `?`, `|`, `-`, `<`, `>`, `=`, `!`, `%`, `@`
255 | | /// Strings containing any of the following characters must be quoted.
256 | | /// `{`, `}`, `\[`, `\]`, `,`, `#`, `\``
    | |________________________________________^
    |
    = help: a backtick may be missing a pair
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/
index.html#doc_markdown
note: the lint level is defined here
   --> insta/src/lib.rs:1:9
    |
1   | #![warn(clippy::doc_markdown)]
    |         ^^^^^^^^^^^^^^^^^^^^
    ```

File: insta/src/macros.rs (hunk 7 above)
Comment (max-sixty) [reply insta/src/macros.rs]: Weirdly I can't repro this...

Comment (CommanderStorm) [reply insta/src/macros.rs]: Here are details to reproduce :

<details><summary><code>rustup --version<code><summary>

```
rustup 1.27.1 (54dd3d00f 2024-04-24)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.81.0 (eeb90cda1 2024-09-04)`
```

<details>

<details><summary><code>git diff | cat<code><summary>

```
diff --git a/insta/src/macros.rs b/insta/src/macros.rs
index 45434b9..9c69005 100644
--- a/insta/src/macros.rs
+++ b/insta/src/macros.rs
@@ -81,7 +81,7 @@ macro_rules! assert_toml_snapshot {
 /// The value needs to implement the [`serde::Serialize`] trait and the snapshot
 /// will be serialized in YAML format.  This does mean that unlike the debug
 /// snapshot variant the type of the value does not appear in the output.
-/// You can however use the [`assert_ron_snapshot!`](crate::assert_ron_snapshot!) macro to dump out
+/// You can however use the [`assert_ron_snapshot!`] macro to dump out
 /// the value in [RON](https://github.com/ron-rs/ron/) format which retains some
 /// type information for more accurate comparisons.
 ///
```

<details>

<details><summary><code>RUSTC_BOOTSTRAP=1 RUSTDOCFLAGS="--cfg=docsrs" cargo doc --no-deps --all-features --keep-going<code><summary>

```
    Checking insta v1.41.0 (/tmp/insta/insta)
 Documenting insta v1.41.0 (/tmp/insta/insta)
 Documenting cargo-insta v1.41.0 (/tmp/insta/cargo-insta)
warning: unresolved link to `assert_ron_snapshot`
   --> insta/src/macros.rs:77:1
    |
77  | / /// Asserts a [`serde::Serialize`] snapshot in YAML format.
78  | | ///
79  | | /// **Feature:** `yaml`
80  | | ///
...   |
120 | | ///
121 | | /// The snapshot name is optional but can be provided as first argument.
    | |________________________________________________________________________^
    |
    = note: the link appears in this line:
            
            You can however use the [`assert_ron_snapshot!`] macro to dump out
                                     ^^^^^^^^^^^^^^^^^^^^^^
    = note: no item named `assert_ron_snapshot` in scope
    = note: `macro_rules` named `assert_ron_snapshot` exists in this crate, but it is not in scope at this link's location
note: the lint level is defined here
   --> insta/src/lib.rs:2:9
    |
2   | #![warn(rustdoc::all)]
    |         ^^^^^^^^^^^^
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` implied by `#[warn(rustdoc::all)]`

warning: `insta` (lib doc) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.19s
   Generated /tmp/insta/target/doc/cargo_insta/index.html and 1 other file
```

<details>

CommanderStorm:
> but I do get this error

Sorry, that was a botched merge from master.
Unsure if we should keep `clippy::doc_markdown`. Seems pretty brittle..

Apparently, the unbalanced backticks part does not properly ignore escaped backticks..

insta/src/macros.rs
max-sixty: Ah, sorry, I hadn't realized this was with changed code

max-sixty:
> Unsure if we should keep `clippy::doc_markdown`. Seems pretty brittle..

As you wish. Definitely worth being liberal on the vendored code... 

The other lint seems great fwiw, has fixed lots of issues which are now ratcheted in.

---

Will merge! Thank you for these @CommanderStorm

Issue #625, opened 2024-09-30
`cargo-insta` v1.40.0 `--workspace-root` regression


ilslv:
### What happened?

Reproduction repo: https://github.com/ilslv/cargo-insta-workspace-root-bug

Hi! I have a quite strange crate layout: workspace with a library that has `[[test]]` section pointing to a test in the workspace root outside the library itself (see the reproduction repo for exact layout and `reproduce.sh` for the steps). This outer test uses `insta` and since 1.40.0 handling of `cargo insta <accept/review/pending-snapshots> --workspace-root tests` stopped seeing snapshots.

Possibly related to the https://github.com/mitsuhiko/insta/issues/593

### Insta Version

1.40.0

max-sixty:
Thank you very much for the reproduction! I will have a look at this very shortly

max-sixty:
Quick summary of the issue:

- Part of the change in 1.40.0 was to handle each package within a workspace separately, rather than an arbitrary mix of both per-workspace and per-package. This fixed a pretty bad bug that had existed for a while: https://github.com/mitsuhiko/insta/issues/396, and so overall was a good change.
- But a side-effect was that scanning for pending snapshots outside of a cargo workspace doesn't work; we scan each package's cargo manifest path: https://github.com/mitsuhiko/insta/blob/a2e70ae221df0f52f1d0b63c3f49b2b580d99316/cargo-insta/src/cli.rs#L453-L460. The workspace root is only used for finding the cargo manifest.
- This indeed comes from the same group of changes which caused #593. #593 comes from insta now inheriting `cargo`s behavior around default packages — maybe a surprising change, but not necessarily a bad state (hence the open question there on whether to change insta's defaults). 
- In contrast, this behavior is fundamentally wrong when there are files outside the path of the `Cargo.toml`
- I think the fix is to look at the `src_path` of the package itself when scanning for pending snapshots, rather than assume it's underneath the path of `Cargo.toml`. (Currently we do that correctly when _writing_ the pending snapshot... this is another area we should unify the logic rather than having similar-but-different logic around the library)
- Another effect of this is that `cargo insta test` will report success here, because it doesn't find the pending snapshots it wrote, which is obv bad.
- One note: this shouldn't actually require passing `--workspace-root` — this is in the cargo manifest and so we should pick that up automatically.

Will try and get to this over the next few days.

PR #626, opened 2024-10-01
Canonicalize snapshot paths before reducing

max-sixty:
Fixes https://github.com/mitsuhiko/insta/issues/625

As ever, the reason is a bit different from the one I initially understood. It's actually that we're reducing the paths by whether one prefixes the other. But in this case the path is only a prefix before it's canonicalized!

On the upside, this case will now work without `--workspace-root`.


max-sixty:
(requires a test before merging)

PR #633, opened 2024-10-03
Redo integration test output

max-sixty:
Much more coherent — shows output of inner tests consistently regardless of how they fail, with a nice prefix. Removes the clunky `assert_success` function.


decathorpe:
Is there a reason why this uses an obsolete os_pipe version (`^0.9`, last release in this branch was 4 years ago) instead of the current os_pipe `^1`?

max-sixty:
Possibly MSRV. Would happily take a PR to upgrade assuming it's compatible (tests will catch it)

decathorpe:
Hm, I don't think so. os_pipe 0.9 doesn't specify MSRV in its metadata, and  os_pipe 1.2.0+ specifies an MSRV of 1.63 ... but cargo-insta already has an MSRV of 1.65.

PR #648, opened 2024-10-07
Add warning about stability to doc comment for internals module

lasernoises:
As discussed in #610 it probably makes sense to be explicit about the `internals` module not being guaranteed to be stable.

I also noticed that `TextSnapshotKind` gets exposed globally. This was added in #581 when it was still called `SnapshotKind`. I wonder if this is necessary to expose toplevel.


max-sixty:
Thanks @lasernoises ! 

I tagged @mitsuhiko in case he has any views, will merge now and we can make changes later if needed.

Issue #492, opened 2024-05-20
Reconsider when to clear additional `.snap.new` files


max-sixty:
Currently we can get the following sequence:

- Process 1: `cargo insta test --accept -- test_foo` is runs on every file change (e.g. with `watchexec`). `test_foo` passes fine.
- Process 2: `cargo insta test --accept` is run manually. This finds an incorrect snapshot and creates a `.snap.new` file
- Process 1: triggered by the file change, it re-runs its command on `test_foo`,  `test_foo` passes fine again. It doesn't accept the `.snap.new` file because it's for an unrelated change. But it does remove it.
- Process 2: completes, and would accept the change in `.snap.new`, but the file is gone. It doesn't raise an error.

So because we have two processes running, we've managed to run `cargo insta test --accept`, missed an incorrect snapshot, and yet every process has reported success.

Should we instead only remove additional `.snap.new` files when running `cargo insta review`? Unlike `cargo insta test --accept`, that's running over every `.snap.new` file; it can't delete ones that relate to other tests. `cargo insta test --accept` can still remove `.snap.new` files for the tests it's running, but wouldn't for other tests.

max-sixty:
Fixed by #651

PR #651, opened 2024-10-09
Don't clear unrelated pending snapshots

max-sixty:
For discussion, re #492.

There are tests which confirm that a passing test removes any associated pending snapshot. And a test that checks `--unreferenced=delete` picks up pending snapshots; so running all tests with `--unreferenced=delete` is a catch-all way of removing any pending snapshots.

My only reservation is whether I'm missing something? Otherwise the proposed behavior seems better.


max-sixty:
Merging, as it's well tested and positive-expected-value

...and am back online more now, so can revert / iterate if it doesn't work as well as I hoped

PR #668, opened 2024-10-13
Add pending deprecation warning for `--accept-unseen`

max-sixty:
Open to this being shot down. If others tend to agree, this is a low-cost way of figuring out whether anyone is using it


max-sixty:
Merging this! As per the issue, we can revert and fix the feature if we get feedback on the feature

PR #672, opened 2024-10-21
chore: updating cargo lock for ron/toml

Freyskeyd:
This PR updates `toml` dependency to avoid discrepancy between the latest `toml` version and the one in `insta`.

When refactoring some part of the configuration on one of my project that is using `toml`, I discovered that my configuration was properly serialized when using `toml = 0.8.x` while when using in `insta` it was failing with`UnsupportedType`. After digging a bit, it is related to [this](https://github.com/toml-rs/toml/issues/553).

To reproduce on a fresh `cargo new --lib` :

`Cargo.toml`
```toml
[dependencies]
serde = { version = "1.0.210", features = ["derive"] }
toml_5 = { package = "toml", version = "0.5" }
toml_8 = { package = "toml", version = "0.8" }
```

`lib.rs`
```rust
use serde::Serialize;

#[derive(Serialize)]
enum SomeType {
    Type1 {},
    Type2 {},
}

impl Default for SomeType {
    fn default() -> Self {
        SomeType::Type1 {}
    }
}

#[derive(Serialize, Default)]
struct Config {
    #[serde(default)]
    some_type: SomeType,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    #[should_panic]
    fn toml_5_test_serialize_config_fails() {
        let cfg = Config::default();

        toml_5::to_string_pretty(&cfg).unwrap();
    }

    #[test]
    fn toml_8_test_serialize_config_ok() {
        let cfg = Config::default();

        toml_8::to_string_pretty(&cfg).unwrap();
    }
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -253,6 +253,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
 
 [[package]]
+name = "equivalent"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
+
+[[package]]
 name = "errno"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -323,6 +329,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "hashbrown"
+version = "0.15.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e087f84d4f86bf4b218b927129862374b72199ae7d8657835f1e89000eea4fb"
+
+[[package]]
 name = "heck"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -350,6 +362,16 @@ dependencies = [
  "thread_local",
  "walkdir",
  "winapi-util",
+]
+
+[[package]]
+name = "indexmap"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "707907fe3c25f5424cce2cb7e1cbcafee6bdbe735ca90ef77c29e84591e5b9da"
+dependencies = [
+ "equivalent",
+ "hashbrown",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


max-sixty:
IIUC, upgrading toml has backward compatibility concerns: https://github.com/mitsuhiko/insta/issues/439.

Is there a version that fixes the issues you were hitting without breaking backward-compat?

Freyskeyd:
Hello, I think that `0.6` were fixing the problem, I can try.

Freyskeyd:
After trying, it is only solved on 0.8.

But I'm a little confused about this, the failing test:

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: toml_inline
Source: insta/tests/test_inline.rs:162
────────────────────────────────────────────────────────────────────────────────
Expression: User {
    id: 42,
    username: "peter-doe".into(),
    email: Email("peter@doe.invalid".into()),
}
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    0     0 │ id = 42
    1       │-username = 'peter-doe'
    2       │-email = 'peter@doe.invalid'
          1 │+username = "peter-doe"
          2 │+email = "peter@doe.invalid"
────────────┴───────────────────────────────────────────────────────────────────
```

Should have basic string, not literal, right?
Literal would be used in place of basic string if needed, like if I use:

```rust
User {
    id: 42,
    username: "peter-doe".into(),
	display_name: r#"peter "The Machine" doe"#.into(),
    email: Email("peter@doe.invalid".into()),
}
```

which would produce:

```toml
id = 42
username = "peter-doe"
display_name = 'peter "The Machine" doe'
email = "peter@doe.invalid"
```


I understand that it can break some existing snapshot, but on the other side, it makes the library unusable for valid types. The only workaround that I found is to preserialize the struct with `toml 0.8`, but it prevents me to use any `redactions` nor `filters`.

max-sixty:
> I understand that it can break some existing snapshot, but on the other side, it makes the library unusable for valid types.

Right, but breaking existing snapshots requires us to bump the major version of insta (which is possible but there's lots of things we'd like to get in; we wouldn't do it for just this).

Is it possible to do this with a feature, like `toml8`? (I haven't tried this sort of thing in rust before, so not sure...)

Freyskeyd:
@max-sixty it means that we will have duplication in `assert_toml_snapshot` with `assert_toml8_snapshot`

max-sixty:
Is it instead possible to do this with a feature?

Freyskeyd:
Due to the fact that we're testing with `--all-features` we could have both `toml` and `toml8` feature at the same time.
Or we need to make a `compile_error` that defines that you can't have `toml` and `toml8` at the same time.
Leading us to remove the `all-feature` flags and having to list every features that we want to enable during tests.

max-sixty:
Yes, we'd need to find some way of it using v8 when both features were enabled. 

To the extent we have both versions of toml on the `toml` feature in your proposal — for the additional macro — are you sure we can't conditionally compile the v5 vs v8 macro as `assert_toml_snapshot` when the `toml` and _not_ the `toml8` feature is enabled?

Or if that's confusing — would be interested to see your proposal above with both dependencies active. I'd have thought the jump for there to conditionally compiling isn't that big (but may well be missing something)

max-sixty:
(I adjusted the title as this isn't a simple chore, feel free to adjust further to the extent this stays open)

max-sixty:
superseeded by #834, thank you!

PR #690, opened 2024-12-10
Don't add `snapshot_kind: text` to snapshots

max-sixty:
This was extremely simple, sorry to leave it hanging.

Will do a self-review tomorrow to confirm I haven't missed anything


nyurik:
horray, thanks!!

max-sixty:
One thing I did miss — this will remove existing `snapshot_kind: text`. Ideally we'd keep those, or it's going to be a lot more churn.

(shows the cost of leaving it for a while — I remember thinking this at the time)

max-sixty:
OK, I realize it's not tractable to have it retain `snapshot_kind: text`. So now making changes to any snapshots (or running `--force-update-snapshots`)  that have `snapshot_kind: text` will remove that metadata item.

I personally disagree with merging this. From the attached issue explaining the original change:

> I appreciate it's some aesthetic churn. But OTOH it only appears when updating snapshots. And for those who care about the cleanliness of diffs, doing a single run of `--force-update-snapshots` will update everything in a single commit. And in general it's important for insta to be able to evolve the snapshot format, albeit in the specific case we could have traded away consistency in metadata to reduce changes.

Merging this adds some additional churn (though admittedly more so since I took a couple weeks to do this).

But I weigh @mitsuhiko 's opinion highly, and appreciate some of the trade for being able to move quickly is occasionally reverting things. So I'll merge now, and release in a few days unless there are any second thoughts.

nyurik:
thx for merging!!! Can't wait to switch to the new version :)

PR #695, opened 2024-12-14
fix(settings): mark drop guard as !Send

### LINKED ISSUE #694: Settings::bind_to_scope()'s drop guard is Send
### What happened?

`Settings::bind_to_scope()`'s drop guard is `Send`, this means you can transfer it to a separate thread and break the settings like that.

This is not normally an issue, but using this in a multi-threaded executor like `tokio` could cause weird failures (when a future changes threads).

### Reproduction steps

1. `let guard = settings.bind_to_scope();`
2. `std::thread::spawn(move || { let _ = guard; })`
3. Observe settings in current thread are not reset.

### Insta Version

1.41.1

### rustc Version

1.83

### What did you expect?

`Settings::bind_to_scope()`'s drop guard should not be `Send` so this is prevented at compile time.

jalil-salame:
Fixes #694

Because the `SettingsDropGuard` modifies a `ThreadLocalKey` (it clones it in the `Settings::bind_to_scope` method, then restores it in its drop impl), it should not be moved to a different thread, as then it would modify that thread's `CURRENT_SETTINGS` instead of the original thread's.

This is not an issue with normal code as the guard is usually unnameable (`let _guard = settings.bind_to_scope();`). but it is a problem with `async` code, as when held across `await` points is can be moved by the runtime to a separate thread.

We fix it by adding a `PhantomData` field to `SettingsDropGuard` that does not implement `Send`. We verify it works by adding a `compile_fail` doctest to the struct that tries to send the drop guard to a different thread.

### DIFF
--- a/insta/src/settings.rs
+++ b/insta/src/settings.rs
@@ -569,7 +569,7 @@ impl Settings {
         CURRENT_SETTINGS.with(|x| {
             let mut x = x.borrow_mut();
             let old = mem::replace(&mut x.inner, self.inner.clone());
-            SettingsBindDropGuard(Some(old))
+            SettingsBindDropGuard(Some(old), std::marker::PhantomData)
         })
     }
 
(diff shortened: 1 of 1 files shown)


jalil-salame:
This is technically a breaking change, but it is a logic error, so I would argue it is not breaking as users relying on this functionality had broken tests already

max-sixty:
I'm not an expert here, but in principle it makes a lot of sense. I'll merge unless @mitsuhiko has any feedback in the next day or so.

@jalil-salame if you'd be up for adding a one-sentence description to the Changelog (+ an attribution if you'd like), that would be great!

PR #698, opened 2025-01-04
Replace `lazy_static` with `once_cell`

FalkWoldmann:
Hi, since both crates are now on MSRV >= 1.60, we can now use `once_cell` over `lazy_static` [0]

[0] https://github.com/matklad/once_cell/blob/4fbd4a53509ca59c8e09efc2ab0a22f21de70f7e/Cargo.toml#L7

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target
 .vscode/*.log
 .aider*
 .env
+.idea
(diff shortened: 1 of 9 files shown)


max-sixty:
Thank you!

Could we fix the lints?

max-sixty:
Thanks @FalkWoldmann !

PR #716, opened 2025-02-01
Don't escape snapshots containing tab and ESC control characters

### LINKED ISSUE #715: Newline gets escaped in `assert_snapshot` if the string contains '\t'
### What happened?

#713 changes the behaviour of `assert_snapshot` so that whenever there are control characters in the string, it gets printed using `format!` which will instead of literally.
When the string contains just `\n` and `\t`, this makes the diff [less readable](https://github.com/jj-vcs/jj/pull/5540/files/01502cb2ac687453cdf4db8e7bc373ab00afedae#r1938256985):
```rs
insta::assert_snapshot!(stdout, @"immutable_bookmark\timmutable\nmutable_bookmark\tmutable\nk\tworking_copy\ny\tmutable\nq\timmutable\nzq\tremote_commit\nzz\t(no description set)\nremote_bookmark@origin\tremote_commit\nalias_with_newline\t roots(\nsiblings\t@-+ ~@");
```

This could be fixed by treating `\t` as `\n` in `has_control_chars`:
https://github.com/mitsuhiko/insta/pull/713/files#diff-2666420e8389f1d842ce108018b05f3f68fa4b920f8c1793a3d4d01f21443544R696-R698

---

Personally I would go even further and prefer the old handling for  `\x1b` (used for ANSI escape codes) as well.
Here's the diff for applying insta 1.42.1 on [jj](https://github.com/jj-vcs/jj) https://github.com/jakobhellermann/jj/commit/7fdb9784e27f02aa9d676b16c09f3f400240fc95 and in my opinion it is much less readable.

<details>

in vscode

in vim
</details>

### Reproduction steps

_No response_

### Insta Version

1.42.1

### rustc Version

rustc 1.85.0-nightly (45d11e51b 2025-01-01)

### What did you expect?

_No response_

jakobhellermann:
fixes #715

`'\0'.is_control() == true`, so I removed that special case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to insta and cargo-insta are documented here.
 
+## 1.42.2
+
+- Stop `\t` and `\x1b` (ANSI color escape) from causing snapshots to be escaped.  #715
+
 ## 1.42.1
 
 - Improved handling of control characters in inline snapshots.  #713
--- a/insta/src/snapshot.rs
+++ b/insta/src/snapshot.rs
@@ -691,11 +691,10 @@ impl TextSnapshotContents {
 
         // Some characters can't be escaped in a raw string literal, so we need
         // to escape the string if it contains them. We prefer escaping control
-        // characters which except for newlines, which we prefer to see as
-        // actual newlines.
+        // characters except for newlines, tabs, and ESC.
         let has_control_chars = contents
             .chars()
-            .any(|c| c != '\n' && c.is_control() || c == '\0');
+            .any(|c| c.is_control() && !['\n', '\t', '\x1b'].contains(&c));
 
         // We prefer raw strings for strings containing a quote or an escape
         // character, and for strings containing newlines (which reduces diffs).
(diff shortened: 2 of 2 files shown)


jakobhellermann:
If you don't want to keep `ESC` as is and render `\u{1b}` instead, I would suggest at least still allowing newlines and escaping each individual line, i.e.

```
        \u{1b}[38;5;0m black \u{1b}[39m
        \u{1b}[38;5;1m red \u{1b}[39m
        \u{1b}[38;5;2m green \u{1b}[39m
        \u{1b}[38;5;3m yellow \u{1b}[39m
```
over
```
\u{1b}[38;5;0m black \u{1b}[39m\n\u{1b}[38;5;1m red \u{1b}[39m\n\u{1b}[38;5;2m green \u{1b}[39m\n\u{1b}[38;5;3m yellow \u{1b}[39m\n
```

max-sixty:
Great, agree, thank you!

Would you mind adding a quick whatsnew?

Would you like a release or no rush?

jakobhellermann:
> Edit: actually are you confident about tab characters? Aren't those a bit difficult to see? 

I suppose it depends on whether you want to test the stuff surrounding the tabs, or whether you want to explicitly check for the existence of tabs over e.g. spaces.

In my use case, I want to test that clap emits certain completions. The formatting is not too important, so it is nice to have the output be readable and use literal tabs:
image

Compare that to `\t`
image

max-sixty:
Yes good point! Agree. And folks can always change it to use the other form manually if they prefer.

(For any future travelers — if this is much worse for some use cases, please post them here and we can make the tradeoff.)

Issue #712, opened 2025-01-23
Multiple `assert_snapshot` within a `allow_duplicates` doesn't work well


ilyagr:
### What happened?

I think there might be several problems here

### Reproduction steps

1. Clone https://github.com/jj-vcs/jj/commit/35440ce1bd62c1752a13f1e4e6353eae783ad666
2. Observe that `cargo nextest run` passes
3. Run `cargo insta test  --workspace --test-runner nextest --force-update-snapshots`
4. Try `cargo nextest run` again

Expected results:

tests pass

Actual result:

Compilation fails

```
error: bare CR not allowed in string, use `\r` instead
   --> cli/src/git_util.rs:820:74
    |
820 |         assert_snapshot!(update(Duration::from_millis(1), 0.10), @"␛[?25l␍ 10% [█▊                ]␛[K...
    |                                                                          ^
    |
help: escape the character
    |
820 |         assert_snapshot!(update(Duration::from_millis(1), 0.10), @"␛[?25l\r 10% [█▊                ]␛[K");
    |                                                                          ++

error: could not compile `jj-cli` (lib) due to 1 previous error
```

Furthermore, these `\r`-s are not the only problem. If that one diff is reverted (`\r` replaced with `␍`again) in https://github.com/jj-vcs/jj/pull/5441, there are more compilation errors, see https://github.com/jj-vcs/jj/actions/runs/12934546007/job/36075816189?pr=5441 . For example,

```
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `""`
   --> cli/tests/test_git_fetch.rs:334:10
    |
334 |     ");""""
    |          ^^ expected one of `.`, `;`, `?`, `}`, or an operator
```

### Insta Version

1.42.0, same behavior with 1.41.1

### rustc Version

rustc 1.86.0-nightly (48a426eca 2025-01-12)

### What did you expect?

Tests to pass.

max-sixty:
Thanks, this seems important.

Do you know whether this replicates without `nextest`? Has it been an issue for a while or are there some new test outputs in `jj`?

ilyagr:
I just started investigating, I think there are at least two problems (as seen by the two error messages).

Running `insta` without `nextest` gives the same result.

I randomly tried running `insta` on commit dddba46 from 6 months ago, and there were no problems. I tried a commit from earlier today, and got the same error. I might try bisecting to find when the problem occurred.

max-sixty:
[...]

PR #722, opened 2025-02-12
Fix inline snapshot handling within `allow_duplicates!` block

yuja:
#712

There are two problems:
1. single-line snapshot isn't deduplicated
2. the last assertion in the `allow_duplicates!` block is always chosen

```rust
#[test]
fn test_dup() {
    for _ in 0..2 {
        insta::allow_duplicates! {
            // shouldn't be updated twice
            insta::assert_snapshot!("foo", @"");
        }
    }
}
```

```rust
#[test]
fn test_bad_location() {
    for _ in 0..2 {
        insta::allow_duplicates! {
            // each snapshot should be updated accordingly
            insta::assert_snapshot!("1", @"
            1a
            1b
            ");
            insta::assert_snapshot!("2", @"
            2a
            2b
            ");
        }
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to insta and cargo-insta are documented here.
 ## 1.42.2
 
 - Stop `\t` and `\x1b` (ANSI color escape) from causing snapshots to be escaped.  #715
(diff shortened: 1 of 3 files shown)


ilyagr:
Passes all of my tests, nice!

max-sixty:
This looks very good, thank you @yuja !

PR #723, opened 2025-02-12
Fix crash on scanning indentation of arbitrary lines

yuja:
```
thread 'main' panicked at cargo-insta/src/inline.rs:130:22:
byte index 32 is out of bounds of `    insta::assert_snapshot!(`
```

### DIFF
--- a/cargo-insta/src/inline.rs
+++ b/cargo-insta/src/inline.rs
@@ -5,7 +5,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use insta::_cargo_insta_support::TextSnapshotContents;
-use proc_macro2::TokenTree;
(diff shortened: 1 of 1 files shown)


ilyagr:
I'm a bit confused about this PR. It conflicts with #722 and my naive attempt to resolve the conflict resulted in broken tests.

Perhaps it should be a draft until #722 is merged?

yuja:
I can rebase this or #722 as needed. They're different patches and I think this bug is more important than #722 because it's a regression.

max-sixty:
Thanks a lot @yuja !

If you want to add a changelog + attribution later, that would be great! No obligation though

yuja:
The bug doesn't exist in the released version, so I don't think changelog entry is needed. Thanks for reviewing.

PR #726, opened 2025-02-14
Allow setting INSTA_WORKSPACE_ROOT at compile time

### LINKED ISSUE #725: Respect INSTA_WORKSPACE_ROOT at compile time
## Problem description

I'm using insta in a rust project that we build using [Yocto](https://www.yoctoproject.org/).

Yocto does some strict checks on binaries to ensure reproducibility and pollution by the host system. So, it will produce warnings if the final binary contains references to the build directory of the host.

In my case also build and package all our tests as binaries in order to run them on a target device. This works and I set the `INSTA_WORKSPACE_ROOT` variable to the correct location on the target file system.

Now the problem I'm facing is that the binary still contains a reference to the build directory if any of the `insta::assert_*` macros is used.

I think this i because the macro by default uses a reference to `CARGO_MANIFEST_DIR` in https://github.com/mitsuhiko/insta/blob/master/insta/src/macros.rs#L27

## Proposal

If `INSTA_WORKSPACE_ROOT` is set at compile time use it instead of `CARGO_MANIFEST_DIR` as the default in the macro.
This way the final binary would not contain the reference to the build directory.

bachp:
This is useful if you already know the intended target location at compile time. For example when installed as a package in a linux distro.

It also avoids some reproducible build issues by avoiding embedding the value of CARGO_MANIFEST_DIR in the final binary.

Fixes #725


insta/src/macros.rs
```diff
@@ -19,12 +19,19 @@ macro_rules! _function_name {
 #[macro_export]
 macro_rules! _get_workspace_root {
     () => {{
-        use std::env;
+        use std::{env, option_env};
 
-        // Note the `env!("CARGO_MANIFEST_DIR")` needs to be in the macro (in
+        // Note the `option_env!("INSTA_WORKSPACE_ROOT")` and `env!("CARGO_MANIFEST_DIR")` needs to be in the macro (in
```
max-sixty, line 24: But is this accurate? I don't think it needs to be in a macro?

And aren't we also referencing `INSTA_WORKSPACE_ROOT` in code above at complie time?

max-sixty (commented):
Looks good directionally; left a question. There's also a `dbg` lingering macro

Thank you!

insta/src/macros.rs
bachp: One reason for doing this is to avoid embedding `CARGO_MANIFEST_DIR` in the produced binary, as this is build machine dependend and thus non repoducible.

If I don't do this, I would assume the string would still end up in the binary.

Maybe a different apporach would more suitable, maybe via conditional compile.

max-sixty: Ah yes, I see.

But let's make the comment accurate (almost paste your reply into the code comment)

max-sixty, line 24:
```suggestion
        // Note the `env!("CARGO_MANIFEST_DIR")` needs to be in the macro (in
```

insta/src/macros.rs
```diff
@@ -19,12 +19,19 @@ macro_rules! _function_name {
...
         // contrast to a function in insta) because the macro needs to capture
         // the value in the caller library, an exclusive property of macros.
-        $crate::_macro_support::get_cargo_workspace(env!("CARGO_MANIFEST_DIR"))
+        // By default the `CARGO_MANIFEST_DIR` environment variable is used as the workspace root.
+        // If the `INSTA_WORKSPACE_ROOT` environment variable is set at compile time it will override the default.
```
max-sixty, line 28:
```suggestion
        // If the `INSTA_WORKSPACE_ROOT` environment variable is set at compile time it will override the default.
        // This can be useful to avoid including local paths in the binary.
```

insta/src/macros.rs
max-sixty: OK, that makes sense; I suggested a change to the comments to ensure they remain accurate. Lmk if I'm mistaken!

bachp: I was also thinking if I could add an actual test that checks if the binary contains a reference to `CARGO_MANIFEST_DIR`.

max-sixty:
All good once we pass the windows test...

insta/src/macros.rs
bachp: I added a test that checks the binary for references to to the `CARGO_MANIFEST_DIR`

bachp:
@max-sixty While working on the tests I came up with a different solution to pass the the compile time `INSTA_WORKSPACE_ROOT`. It's not done the same way as the `CARGO_MANIFEST_DIR`.

I used an enum to convey the information if the workspace should be used as is, or if detection via `cargo metadata` is required.

Let me know which variant you prefer, it's explicitly done as a separate commit.

cargo-insta/tests/functional/workspace.rs
```diff
@@ -465,6 +490,138 @@ fn test_snapshot() {
...
+        .args(["test", "--no-run"])
+        .env("INSTA_WORKSPACE_ROOT", moved_workspace_compile.path())
+        .current_dir(&test_project.workspace_dir)
+        .output()
+        .unwrap();
+    assert!(&output.status.success());
+
+    let test_binary_path = find_test_binary(&test_project.workspace_dir);
+
+    let config = rust_strings::FileConfig::new(&test_binary_path).with_min_length(5);
```
max-sixty, line 535: What does this do?

cargo-insta/tests/functional/workspace.rs
```diff
@@ -465,6 +490,138 @@ fn test_snapshot() {
...
+    let extracted_strings =
+        rust_strings::strings(&config).expect("Unable to extract strings from binary");
+
+    assert_eq!(
+        extracted_strings
+            .iter()
+            .filter(|(s, _)| s.contains(test_project.workspace_dir.to_str().unwrap()))
+            .count(),
+        1,
+        "The final binary contains not the expected number of references to the cargo workspace"
```
max-sixty, line 545: Either here or in a comment, can we answer why this should be 1? Same in the next statement

insta/src/env.rs
```diff
@@ -430,18 +430,35 @@ pub fn snapshot_update_behavior(tool_config: &ToolConfig, unseen: bool) -> Snaps
     }
 }
 
+pub enum Workspace {
+    DetectWithCargo(&'static str),
+    UseAsIs(&'static str),
```
max-sixty, line 435: Yes, this seems good!

max-sixty: (Maybe we could change the names to something more literal / nouns, but it's fine atm)

max-sixty (approved):
Approach looks good!

We could make things a bit more descriptive, to try and make it more maintainable, but also not totally obligatory. Will merge on the next iteration!

cargo-insta/tests/functional/workspace.rs
```diff
@@ -340,6 +340,52 @@ fn test_virtual_manifest_single_crate() {
...
+                    && entry.metadata().map(|m| m.is_file()).unwrap_or(false) // Is a file, not a directory
+        })
+        .map(|entry| entry.path())
+        .expect("Failed to find test binary")
+}
+
+// Can't get the test binary discovery to work on Windows, don't have a windows
+// machine to hand, others are welcome to fix it. (No specific reason to think
+// that insta doesn't work on windows, just that the test doesn't work.)
+#[cfg(not(target_os = "windows"))]
```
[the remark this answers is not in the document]
bachp: The problem is that cargo or clippy produce an error on windows as the function is never used there. Maybe it would be better to just do `#[allow(dead_code)]`.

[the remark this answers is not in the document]
bachp: I changed it to allow dead code and the comment is now only on the test case itself

cargo-insta/tests/functional/workspace.rs
```diff
@@ -465,6 +490,138 @@ fn test_snapshot() {
     .success());
 }
 
+// Can't get the test binary discovery to work on Windows, don't have a windows
+// machine to hand, others are welcome to fix it. (No specific reason to think
+// that insta doesn't work on windows, just that the test doesn't work.)
```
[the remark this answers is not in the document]
bachp: Yes, removed it same as aboce

cargo-insta/tests/functional/workspace.rs
bachp: I no strip the binaries when building so that it doesn't contain the debug symbols.

This way the test becomes more clear as I don't need to account for the additional reference in the debug symbols.

I also made the message of the assert more explicit of what is actually expected.

max-sixty:
Great, thanks @bachp !

Merging to ensure it goes in. If you want to follow-up with an addition to the changelog, that'd be welcome; but no obligation.

bachp:
@max-sixty Thanks.

I wasn't sure how you handle changelog entries. But I think it's worth an entry there.

PR #729, opened 2025-02-25
Qualify all references in macros to avoid name clashes

aschey:
This change replaces all unqualified references in macros with a qualified reference from `_macro_support` to avoid name clashes. I ran into the same issue as this https://github.com/tokio-rs/tracing/issues/3023 with the `line!` macro.

I went ahead and used a qualified path for any references to `core` and `std` out of an abundance of caution since some people do have crates named `core` (https://github.com/tokio-rs/tracing/issues/2761).


max-sixty:
Many thanks @aschey !

PR #747, opened 2025-03-27
Bump syn to latest version 2.0.100

yuja:
This fixes parsing of `+ use<'a>` syntax.

"syn" claims to support rustc 1.61, so I think it's compatible with MSRV of cargo-insta.


max-sixty:
thanks!

PR #748, opened 2025-03-28
Minor optimization of required hashes

nyurik:
slightly cleaner and faster (no clone), and no panics


max-sixty:
much nicer, thanks!

PR #789, opened 2025-07-23
Update console dependency from 0.15.4 to 0.16.0

musicinmybrain:
The only breaking change in [`console` 0.16.0](https://github.com/console-rs/console/releases/tag/0.16.0) is that crates that depend on `console` with `default-features = False` may need to explicitly enable the new `std` feature. It looks like we do need to do that in the `insta` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,7 +70,7 @@ version = "1.43.1"
 dependencies = [
  "cargo_metadata",
  "clap",
- "console",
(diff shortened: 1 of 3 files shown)


musicinmybrain:
It looks like `console` 0.16.0 requires Rust 1.66, so this is blocked on MSRV, and this will have to be a downstream-only patch in Fedora for now.

PR #796, opened 2025-08-30
Allow passing args through to test binary with nextest

max-sixty:
Pass `cargo insta test -- --status-level fail -- --nocapture`, with an additional `--` for args to nextest.

Currently it maintains backward compat -- a single `--` will pass to the test binary only, with a depreciation.

(for transparency, Claude Code helped with the code)

### DIFF
--- a/cargo-insta/src/cli.rs
+++ b/cargo-insta/src/cli.rs
@@ -119,7 +119,7 @@ struct ProcessCommand {
     quiet: bool,
 }
 
-#[derive(Args, Debug)]
+#[derive(Args, Debug, Clone)]
 #[command(rename_all = "kebab-case", next_help_heading = "Test Runner Options")]
 struct TestRunnerOptions {
     /// Test only this package's library unit tests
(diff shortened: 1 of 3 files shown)


glehmann:
Thanks for working on this! I'm not sure the double `--` would make the best user experience though.
Maybe it would be better to copy the relevant nextest options in `cargo insta test`? Or add a `--test-runner-options`?

max-sixty:
> Or add a `--test-runner-options`?

I considered this. but then would `--test-runner-options="--status-level fail -- --nocapture"` give the same result as `cargo insta test -- --status-level fail -- --nocapture` would in the proposed code?

(and then less keen on adding options to insta to track all of nextest's)

glehmann:
> > Or add a `--test-runner-options`?
> 
> I considered this. but then would `--test-runner-options="--status-level fail -- --nocapture"` give the same result as `cargo insta test -- --status-level fail -- --nocapture` would in the proposed code?

I was thinking of `cargo insta --test-runner nextest --test-runner-options="--status-level=fail" -- --no-capture`, which continues to use `--` as currently.

But both examples you're showing are good enough IMO. I would be more uncomfortable with `cargo insta test -- -- --no-capture`. That's a lot of `--`!

Anyway, that's probably a matter of taste. In the end, any solution would work for me :+1:

max-sixty:
on reflection, this code is overly complicated and would be much simpler if we didn't run doctests separately. so I got around to doing https://github.com/mitsuhiko/insta/pull/803, and we can merge a version of this when we've removed the supplemental doctest run

PR #798, opened 2025-09-02
don't panic when getting the cargo workspace

adriangb:
This fixes https://github.com/mitsuhiko/insta/discussions/794.
I unfortunately do not have an MRE (I tried, a lot), but it seems reasonable to not panic here.


adriangb:
Thanks, done!

adriangb:
CI failure seems unrelated to this change? https://github.com/mitsuhiko/insta/actions/runs/17435656048/job/49513973977?pr=798#step:5:178

insta/src/env.rs
```diff
@@ -459,51 +459,80 @@ pub fn get_cargo_workspace(workspace: Workspace) -> Arc<PathBuf> {
...
-                )
-            })
+        .or_insert_with(|| match get_cargo_workspace_inner(manifest_dir) {
+            Some(path) => path,
+            None => {
+                eprintln!(
+                    "error: failed to get cargo workspace for {}, will use manifest directory",
+                    manifest_dir
+                );
+                Arc::new(PathBuf::from(manifest_dir))
```
adriangb, line 474: Btw I'm not sure what this return means. Everything seems to work for us despite hitting this error, but I'm not sure what the intention of this function was in the first place, I somewhat guessed that this is the right value to return under error conditions...

max-sixty:
can we merge in `main`? I just merged a fix for the clippy issue

adriangb:
done

PR #806, opened 2025-10-07
chore: speed up JSON float rendering

nyurik:
Avoid allocating new String for each float number rendering, and de-dup the code a bit.


max-sixty:
thanks @nyurik !

Issue #805, opened 2025-10-06
Allow "pretty compact json" snapshots


nyurik:
Snapshots are mostly read by humans when they change. Yet, some structs are so big that they result in enormous unreadable snapshots, like [this one](https://github.com/oxibus/can-dbc-pest/blob/49e6f960161723bccb6d63110e802e7f33ed52e6/tests/snapshots-dbc-cantools/BU_BO_REL_.snap).

Usually I try to use `assert_yaml_snapshot` instead of `assert_debug_snapshot` if the struct implements `serde`, but both yaml and json are too verbose:

```yaml
---
source: tests/snapshots.rs
---
pos:
  - 0
  - 14134
pairs:
  - pos:
      - 0
      - 14134
    rule: file
...

```

```json
{
  "pos": [
    0,
    14134
  ],
  "pairs": [
    {
      "pos": [
        0,
        14134
      ],
      "rule": "file"
...
```

## Proposal

A while ago I found that the "pretty-compact" usually works much better. There is [json-pretty-compact crate](https://crates.io/crates/json-pretty-compact) that should produce something like this. Basically it uses space-laced one-liner until it no longer fits on one line, at which point it backtracks to multi-line.  I think the `assert_compact_json_snapshot!` should use it instead of the currently less useful "oneliner or pretty" mode.

```json
{
  "pos": [0, 14134],
  "pairs": [
    {"pos": [0, 14134], "rule": "file"},
...
```

PR #807, opened 2025-10-07
feat: assert_compact_pretty_json_snapshot

nyurik:
implement #805

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -360,6 +360,7 @@ dependencies = [
  "console",
  "csv",
  "globset",
+ "json-pretty-compact",
  "once_cell",
  "pest",
  "pest_derive",
@@ -367,6 +368,7 @@ dependencies = [
  "ron",
  "rustc_version",
  "serde",
+ "serde_json",
  "similar",
  "similar-asserts",
  "toml",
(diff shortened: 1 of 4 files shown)


max-sixty:
I very much agree with the value of compact snapshots

but I'm a bit worried about macro proliferation; that we have a different macro for each combination of settings. I was already a bit hesitant about `assert_compact_json_snapshot`, but IIRC @mitsuhiko was keener.

alternatives would include:
- start with a wrapper in the crate being tested that ran the compact json and then passed to `assert_snapshot`
- adding code to `insta` so we can set settings for how to serialize

@mitsuhiko do you have thoughts?

nyurik:
can we simply re-target `assert_compact_json_snapshot` to use this new renderer?

P.S. the fewer settings we have in insta, the better - conventions over configuration :)

max-sixty:
> can we simply re-target `assert_compact_json_snapshot` to use this new renderer?

but I think it would be backward incompat because it would require enabling a feature? (and we would want a new feature as it would be adding a dependency...)

nyurik:
unless someone wants to re-implement it in-place... Than again, I have no idea why `insta` wants to maintain a JSON rendering code rather than just use serde_json?  Doesn't seem like a high value thing, esp considering that serde_json is looked at by so many more people...

nyurik:
P.S. technically adding a small mandatory dependency is not a biggie of course

max-sixty:
> Than again, I have no idea why `insta` wants to maintain a JSON rendering code rather than just use serde_json?

yeah, there's some modest history there; discussions are easy to grab

PR #810, opened 2025-10-20
insta: Cargo: allow globset version up to 0.4.16

g0hl1n:
## Summary

[globset v0.4.17](https://github.com/BurntSushi/ripgrep/releases/tag/globset-0.4.17) changed its rust edition from 2021 to 2024. As insta is currently on edition 2021 ensure to use globset versions <0.4.17.

## References
- https://github.com/BurntSushi/ripgrep/commit/a60e62d9ac559b1468d4dd27b80a7b662a600e0c
- https://github.com/BurntSushi/ripgrep/pull/3149

### DIFF
--- a/insta/Cargo.toml
+++ b/insta/Cargo.toml
@@ -50,7 +50,7 @@ pest = { version = "2.1.3", optional = true }
 pest_derive = { version = "2.1.0", optional = true }
 ron = { version = "0.7.1", optional = true }
 toml = { version = "0.5.7", optional = true }
-globset = { version = "0.4.6", optional = true }
+globset = { version = ">= 0.4.6, < 0.4.17", optional = true }
 walkdir = { version = "2.3.1", optional = true }
 similar = { version = "2.1.0", features = ["inline"] }
 regex = { version = "1.6.0", default-features = false, optional = true, features = [


max-sixty:
isn't default cargo behavior to allow any patch version?

what's the observable bad outcome with the existing command?

g0hl1n:
> isn't default cargo behavior to allow any patch version?

Exactly. That's the problem IMHO. insta is in this case allowing a dependency which doesn't matches its rust edition.

> what's the observable bad outcome with the existing command?

The problem shows when compiling with a rust toolchain which supports only edition 2021, but doesn't support 2024 (as you do if you build using YoctoProject/OpenEmbedded walnascar for example). If `globset` is updated via `cargo update` it's bumped to 0.4.17. But then the project fails to compile as the compiler doesn't support the rust edition of globset.

IMHO the "root cause" is having `globset` changing its rust edition in a bugfix release (which shouldn't happen if you adhere to SemVer). But as it seems this is somehow a "policy" of ripgrep/globset the only solution would be to either limit the maximum supported globset version or bump the rust edition of insta.

Or have I missed/misinterpreted something?

max-sixty:
but crates can compile with different editions IIUC?

can we see a reproducible example of an error?

g0hl1n:
Sure. Here's the message from the cargo build triggered by bitbake:

```
| Caused by:
|   failed to parse manifest at `/path/to/build/tmp/work/cortexa53-crypto-poky-linux/project/2025.272.0/sources-unpack/cargo_home/bitbake/globset-0.4.17/Cargo.toml`
|
| Caused by:
|   feature `edition2024` is required
|
|   The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.84.0 (66221abde 2024-11-19)).
|   Consider trying a more recent nightly release.
|   See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

max-sixty:
OK yes; I think there's some disagreement in the rust community re whether to do this in libraries vs apps, but no downside on merging this

and the MSRV resolver will solve everything when we're up to that rust version

max-sixty:
thank you @g0hl1n !

PR #817, opened 2025-11-10
feat(vscode-ext): support source opening

Binlogo:
## What does this PR do?

It enables quick opening of the source file in VSCode using CMD+click with the Extension.

### Preview

image

### DIFF
--- a/vscode-insta/src/SnapshotDocumentLinkProvider.ts
+++ b/vscode-insta/src/SnapshotDocumentLinkProvider.ts
@@ -1,0 +1,134 @@
+import * as path from "path";
+import {
+  DocumentLink,
+  DocumentLinkProvider,
+  Position,
+  Range,
+  Selection,
+  TextDocument,
+  Uri,
+  window,
+  workspace,
+} from "vscode";
+
+type SourceLink = {
+  sourcePath: string;
+  range: Range;
+  line?: number;
+};
+
+export const SOURCE_COMMAND = "mitsuhiko.insta.open-source-location";
+
+function extractSourceLink(document: TextDocument): SourceLink | undefined {
+  let sourcePath: string | undefined;
+  let lineNumber: number | undefined;
+  let sourceLineIndex = -1;
+
+  // Simple pass through all lines
+  for (let i = 0; i < document.lineCount; i++) {
+    const text = document.lineAt(i).text;
+    
+    // Find source path
+    if (text.startsWith("source:")) {
+      sourcePath = text.slice(7).trim(); // "source:" is 7 characters
+      sourceLineIndex = i;
+    }
+    
+    // Find assertion line (optional)
+    if (text.startsWith("assertion_line:")) {
+      const lineStr = text.slice(15).trim(); // "assertion_line:" is 15 characters
+      lineNumber = parseInt(lineStr, 10);
+    }
+  }
+
+  if (sourcePath && sourceLineIndex >= 0) {
+    // Create a simple range for the entire source line
+    const line = document.lineAt(sourceLineIndex);
+    const range = new Range(
+      new Position(sourceLineIndex, 0),
+      new Position(sourceLineIndex, line.text.length)
+    );
+    
+    return {
+      sourcePath,
+      range,
+      line: lineNumber,
+    };
+  }
+
+  return undefined;
+}
+
+async function resolveSourceUri(
+  document: TextDocument,
+  sourcePath: string
+): Promise<Uri | undefined> {
+  // Handle absolute paths
+  if (path.isAbsolute(sourcePath)) {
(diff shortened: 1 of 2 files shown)


max-sixty:
thanks for the suggestion!

I don't know this code well, so I asked Claude to take a look. Can we indeed simplify? If not, I trust you more than Claude; but would like to keep this code simple if possible

<details>


⏺ Based on my analysis, here's my review of PR #817:

  Summary

  The PR uses the correct API pattern (DocumentLinkProvider + custom command) since VSCode's DocumentLink doesn't support line/column positions directly. However, the implementation is over-engineered for cases
  that don't exist in insta's actual snapshot format.

  Issues Found

  1. Handles Non-Existent Fields (vscode-insta/src/SnapshotDocumentLinkProvider.ts:84-96)

  The parser handles line: and column: fields that don't exist in insta's metadata structure:

  const lineMatch = text.match(/^(?:assertion_)?line:\s*(\d+)/);  // "line:" doesn't exist
  const columnMatch = text.match(/^column:\s*(\d+)/);              // "column:" doesn't exist

  Reality: Insta only has assertion_line in pending snapshots (see insta/src/snapshot.rs:232-276). No line: or column: fields exist.

  2. Quote Handling Complexity (SnapshotDocumentLinkProvider.ts:45-51)

  The code handles quoted vs unquoted paths, but insta's YAML output doesn't quote source paths:

  const hasQuotes = sourcePath.startsWith("\"") && sourcePath.endsWith("\"");
  if (hasQuotes && sourcePath.length >= 2) {
    sourcePath = sourcePath.slice(1, -1);
    // ... adjust positions ...
  }

  Reality: All snapshot examples show unquoted paths: source: insta/tests/test_basic.rs

  3. Trailing Location Parsing (SnapshotDocumentLinkProvider.ts:54-66)

  Handles source: path/to/file:10:5 format, but insta stores line numbers separately in assertion_line, not in the source path.

  4. Over-Complex Path Resolution (SnapshotDocumentLinkProvider.ts:116-139)

  Tries multiple fallback paths (workspace folder, all workspace folders, document directory) with file existence checks. For a Rust project, snapshots are always relative to the workspace root.

  Simpler Approach

  function extractSourceLink(document: TextDocument): SourceLink | undefined {
    // Parse YAML frontmatter between --- markers
    const text = document.getText();
    const frontmatterMatch = text.match(/^---\n([\s\S]*?)\n---/);
    if (!frontmatterMatch) return undefined;

    const frontmatter = frontmatterMatch[1];

    // Extract source (required) and assertion_line (optional, only in .snap.new files)
    const sourceMatch = frontmatter.match(/^source:\s*(.+)$/m);
    const lineMatch = frontmatter.match(/^assertion_line:\s*(\d+)$/m);

    if (!sourceMatch) return undefined;

    const sourcePath = sourceMatch[1].trim();
    const line = lineMatch ? parseInt(lineMatch[1], 10) : undefined;

    // Calculate range for the clickable link
    const sourceLineStart = text.indexOf(`source: ${sourcePath}`);
    const lineNumber = text.substring(0, sourceLineStart).split('\n').length - 1;
    const startChar = text.substring(sourceLineStart).indexOf(sourcePath);

    return {
      sourcePath,
      line,
      range: new Range(
        new Position(lineNumber, startChar),
        new Position(lineNumber, startChar + sourcePath.length)
      )
    };
  }

  async function resolveSourceUri(
    document: TextDocument,
    sourcePath: string
  ): Promise<Uri | undefined> {
    // Handle absolute paths
    if (path.isAbsolute(sourcePath)) {
      return Uri.file(sourcePath);
    }

    // Resolve relative to workspace root (snapshots are always workspace-relative)
    const workspace = workspace.getWorkspaceFolder(document.uri);
    return workspace ? Uri.file(path.join(workspace.uri.fsPath, sourcePath)) : undefined;
  }

  What's Good

  - ✅ Correct use of command pattern (necessary due to VSCode API limitation)
  - ✅ Good UX with CMD+click to jump to source
  - ✅ Proper handling of both file and inline snapshots

  Recommendations

  Option 1: Simplify to match actual format
  - Remove line:, column: field handling
  - Remove quote handling
  - Remove trailing location parsing
  - Simplify path resolution to workspace-relative only

  Option 2: Keep defensive code
  - If planning future insta features (like column support), document this
  - Add comments explaining why handling non-existent cases

  The command pattern itself is standard and correct—the complexity is in the parser, not the VSCode integration pattern.

</details>

Binlogo:
You're right; the current implementation is a bit complicated. This is my first time contributing to a VSCode extension with Codex's help, and I thought quick open would be useful.

I've simplified the code. Please take a look.

max-sixty:
super, thank you @Binlogo !

PR #819, opened 2025-11-16
chore: update `ron` to 0.12

wetneb:
I made this update as part of updating `ron` in Debian and thought that you might like to have it upstream :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,16 +12,19 @@ dependencies = [
 ]
 
 [[package]]
-name = "base64"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
-
-[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bitflags"
+version = "2.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


max-sixty:
thanks for the contribution! could we try getting tests & the build passing?

wetneb:
Getting the build to pass sounds like an excellent idea :)

Rust was driving me slightly mad in this case, first by complaining about this "unused" import which then becomes "missing" once I remove it! Of course it's because I should have feature-gated it… Sorry about that!

max-sixty:
> first by complaining about this "unused" import which then becomes "missing" once I remove it

lol

max-sixty:
merged; thank you @wetneb !

tmke8:
Huh, I thought insta was never going to update `ron` because it leads to different snapshots because the `ron` format changed? https://github.com/mitsuhiko/insta/pull/414#issuecomment-1780579418

This did cause churn in my snapshots, though I personally don't really mind.

wetneb:
Sorry about that churn, I should have anticipated and highlighted that. I hope it's an acceptable inconvenience if it only happens rarely. Perhaps it would be worth adding a comment in `Cargo.toml` warning against bumping this dependency on a whim.

It would probably be worth highlighting the update in the [release description](https://github.com/mitsuhiko/insta/releases/tag/1.44.1), perhaps.

max-sixty:
let's see whether we get any feedback. I admittedly didn't realize it would hit us, and we don't have the same level of testing for `ron` as we do for others.

I agree we likely can't stay on an old version forever, and the likelihood we get an insta 2.0 together soon is probably not that high

if we get a lot of issues, we could consider reverting / trying to handle both (but IIRC that's quite difficult...)

ilyagr:
Just for reference, here's what the most recent update of `insta` looked like for me:

https://github.com/jj-vcs/jj/commit/2f0132a765518a8df705fd00e10dcc05862c3799

I don't think the changes are ron-related (which makes it less bad to have ron-related or toml-related changes, in my view).

(This are the minimal changes to make tests pass again; `--force-update-snapshots` causes many more changes)

Issue #827, opened 2025-11-24
Please support projects that use `clippy::needless_raw_strings`


joshtriplett:
I like to use the clippy lint `needless_raw_strings`, which regularly catches changes elsewhere in my projects that make a raw string no longer necessary.

I saw mention in another issue that you want to use raw strings to avoid diff noise, and I can appreciate that.

However, for many projects, the output being diffed might *never* need a raw string, and the diff noise is only likely to occur if the output itself noticeably changes to start needing a raw string to represent.

Given that, would it be possible to support `clippy::needless_raw_strings` and generate non-raw strings for strings that contain nothing requiring a raw string?

max-sixty:
@joshtriplett check out #828

joshtriplett:
@max-sixty Thank you! I'm currently testing 1.45, but I'm still seeing it producing raw strings when there are no quotes or backslashes.

If I write the following:

```rust
#[test]
fn test() {
    let result = "a\nb\n";
    insta::assert_snapshot!(result.to_string(), @"");
}
```

Then `insta` turns it into the following:

```rust
#[test]
fn test() {
    let result = "a\nb\n";
    insta::assert_snapshot!(result.to_string(), @r"
    a
    b
    ");
}
```

max-sixty:
would you mind checking the version? the library version of `insta` is the relevant version (I've confused myself with this a couple of times!)

I added a test in #848 with that exact case, and it does seem to remove the `r`

PR #828, opened 2025-11-24
Support clippy::needless_raw_strings lint

max-sixty:
this would fix https://github.com/mitsuhiko/insta/issues/827

@mitsuhiko do you have thoughts? I'm mildly positive on it but will leave it open for a while because I think you voiced weak skepticism in the past.

Claude comment below

---

## Changes

- Modified `to_inline()` to only use raw strings when content contains backslashes or quotes
- Removed newlines from raw string trigger condition
- Added `clippy::needless_raw_strings` lint to the project
- Force-updated all snapshots to use the new logic
- Added allow attribute for vendored YAML code

## Test Results

✅ All tests pass  
✅ Zero clippy warnings in production code  
✅ 44+ needless raw strings converted to regular strings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to insta and cargo-insta are documented here.
 
 ## Unreleased
 
+## 1.44.2
+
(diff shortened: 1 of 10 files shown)


max-sixty:
will merge, but would be backward compat with existing snapshots if @mitsuhiko had strong reservations post

PR #830, opened 2025-11-27
Fix backward compatibility for legacy inline snapshot format

max-sixty:
## Summary

This fixes a regression introduced in 1.44.0 where existing inline snapshots using the legacy multiline format would fail to match.

**The problem:** PR #563 changed how inline snapshots handle leading newlines. This caused snapshots like:

```rust
insta::assert_snapshot!(value, @r"
    Unconflicted Mode(FILE) 0839b2e9412b
    ");
```

to fail when the actual value is `"Unconflicted Mode(FILE) 0839b2e9412b"`.

The legacy format stored single-line content in a multiline raw string with source code indentation. After `from_inline_literal` processing, this becomes `"    Unconflicted...\n    "` — the leading spaces from code indentation weren't being stripped in the legacy comparison path.

**The fix:** In `matches_legacy`, detect the legacy single-line-in-multiline pattern:

```rust
let is_legacy_single_line_in_multiline =
    sc.contents.contains('\n') && sc.contents.trim_end().lines().count() <= 1;
```

When this pattern is detected, apply `trim_start()` to strip the source code indentation artifacts.

This ensures:
- ✅ Legacy snapshots continue to pass (with a warning to update)
- ✅ Modern single-line snapshots with intentional leading spaces are preserved  
- ✅ True multiline snapshots are unaffected

## Test plan

- [x] Added 8 comprehensive functional tests covering:
  - Single-line content in multiline format (the jj bug)
  - Raw string multiline variant
  - Force update reformatting works
  - True multiline content unaffected
  - Multiline with relative indentation preserved
  - Intentional leading spaces correctly fail (no false positives)
  - Trailing whitespace line edge case
  - Empty snapshot edge case
- [x] All 72 functional tests pass
- [x] Full test suite passes

Reported-by: jj-vcs/jj  
See: https://github.com/mitsuhiko/insta/pull/819#issuecomment-3583709431

### DIFF
--- a/cargo-insta/tests/functional/back_compat.rs
+++ b/cargo-insta/tests/functional/back_compat.rs
@@ -1,0 +1,383 @@
+use insta::assert_snapshot;
+use std::process::Stdio;
+
+use crate::TestFiles;
+
+// Tests for backwards compatibility with older snapshot formats.
+//
+// These tests verify that upgrading insta doesn't cause existing valid
+// snapshots to fail. Legacy formats should still match via matches_legacy.
+//
+// See: https://github.com/mitsuhiko/insta/pull/819#issuecomment-3583709431
+
+/// Test that single-line content stored in multiline format still passes.
+///
+/// This reproduces the issue reported at:
+/// https://github.com/jj-vcs/jj/commit/2f0132a765518a8df705fd00e10dcc05862c3799
+#[test]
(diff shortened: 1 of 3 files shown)


max-sixty:
summary done by Claude, as is some of the code...

this fixes the small regression with leading white space on in a multiline snapshots with a single line of text

PR #846, opened 2025-12-18
Skip non-insta snapshot files in unreferenced detection

### LINKED ISSUE #845: Exclude folders from `--unreferenced=reject` checks
In the crates.io codebase we have a mix of Rust and JS code. We use `insta` for a lot of the Rust code, and increasingly `vitest` for the JS code.

We use `cargo insta test --require-full-match --unreferenced=reject` to run the Rust test suite on CI. This worked fine until we tried to use the snapshot feature of `vitest`, which unfortunately also creates `.snap` files. These files now get picked up by `cargo-insta` and reported as unreferenced insta snapshots.

Is there some way to influence what paths are relevant to the "unreferenced snapshots" detection? What I've seen so far are options for gitignored and hidden files, but those unfortunately won't solve our issue.

If there is no such option yet, would there be interest in supporting such a use case? I'm happy to do the implementation work, but we should probably first talk about how this should look like to the enduser :)

max-sixty:
## Summary

- Adds `is_likely_insta_snapshot()` function that reads only 16 bytes to check for `---\nsource:` prefix
- Filters unreferenced snapshot detection to skip files that don't match insta's format
- Adds functional tests for vitest and jest snapshot file preservation

This allows projects using both insta and other snapshot testing tools (like vitest or jest) to use `--unreferenced=reject` without false positives.

Fixes #845

## Test plan

- [x] Added `test_unreferenced_ignores_non_insta_snapshots` - verifies vitest/jest `.snap` files are not deleted by `--unreferenced=delete`
- [x] Added `test_unreferenced_reject_ignores_non_insta_snapshots` - verifies `--unreferenced=reject` doesn't fail on non-insta `.snap` files
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cargo-insta/src/cli.rs
+++ b/cargo-insta/src/cli.rs
@@ -1005,6 +1005,24 @@ fn test_run(mut cmd: TestCommand, color: ColorWhen) -> Result<(), Box<dyn Error>
     }
 }
 
+/// Quick check if a file is likely an insta snapshot by checking for the
+/// `---\nsource:` prefix. This distinguishes insta snapshots from other
+/// snapshot formats (e.g., vitest, jest) that may use the same `.snap` extension.
+///
+/// Only reads 16 bytes for efficiency (longest pattern is 13 bytes for CRLF).
+fn is_likely_insta_snapshot(path: &Path) -> bool {
+    use io::Read;
+    let mut buf = [0u8; 16];
+    let Ok(mut file) = fs::File::open(path) else {
+        return false;
+    };
+    let Ok(n) = file.read(&mut buf) else {
+        return false;
+    };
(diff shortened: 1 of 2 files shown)


max-sixty:
@turbo87 lmk any feedback, ty

max-sixty:
great if you can install and try it out...

Turbo87:
you mean `cargo install` from git? it's not released yet, right?

max-sixty:
yes!

Turbo87:
> great if you can install and try it out...

```
info: no unreferenced snapshots found
```

seems to work perfectly. thanks again! :)

PR #834, opened 2025-11-27
Up-to-date, backwards-compatible TOML

spoutn1k:
Implements the method discussed in #439, fixing it. Rebases #672.

PR #849, opened 2025-12-29
Fix backward compatibility with TOML format produced by insta < 1.45.0

chitoku-k:
This PR fixes backward compatibility with insta < 1.45.0 that has been broken since #834.

As far as I tested, the tests that were added in #834 did not match the behavior before insta 1.45.0 and simply produced new behavior. I tried to fix the implementation as long as test cases and ensured it locally by adding back `toml 0.5.7` and temporarily reverting back to `toml::to_string_pretty(&content)` to test if it matches the original behavior except for those that panicked before, which does not break compatibility at all.


max-sixty:
thank you very much @chitoku-k! I think you're correct; apologies for releasing the backward-incompat code...

PR #854, opened 2026-01-03
Fix path compatibility: try direct strip_prefix first, normalize as fallback

max-sixty:
This is an alternative approach to #852 that should work for both Bazel and Windows.

## Summary
- Try direct `strip_prefix` first (preserves symlinks, works for Bazel)
- Fall back to normalizing with `canonicalize()` if that fails (handles Windows path format differences)

The key insight is: **try direct first, normalize only if that fails**. This is the opposite of the original approach which normalized first (breaking Bazel).

## Test plan
- [x] Local tests pass (93/93)
- [ ] Windows CI passes
- [ ] Bazel integration works (tested previously)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,8 @@ target
 .env
 .idea
 *.rlib
+
+# Bazel artifacts
+BUILD.bazel
+WORKSPACE.bazel
+bazel-*


max-sixty:
Consolidated into #852

Issue #860, opened 2026-01-25
assert_png_snapshot! or similar macros for image snapshots


dstu:
I'm interested in using insta for integration testing of UI code via screenshots. Some desiderata:

1. Snapshots are represented as ordinary image files on disk (presumably in the same way that binary snapshots currently have their content in separate files, rather than encoded in a `.snap` file).
2. Images are compared on the basis of image content (so two images that are compressed differently but decompress to the same image are considered to match, and extraneous metadata fields are ignored).

Binary snapshots almost provide this behavior, but I don't see a straightforward way to get (2). I have looked at a couple of different ways to implement this, and I have a working implementation at https://github.com/dstu/insta/tree/feature/assert_png_snapshot of less than 1000 lines that tries to be minimally disruptive. I haven't opened a PR yet because some discussion about design may be warranted.

The short of it is that this feature branch adds a late-binding pathway for injecting semantics about how to handle a particular kind of `Snapshot` data.

This is done via a trait (which I have called `Comparator`) that provides a function which compares two `Snapshot`s:

```rust
fn matches(&self, config: &ToolConfig, reference: &Snapshot, test: &Snapshot) -> bool;
```

An implementation of `Comparator` is passed to `runtime::assert_snapshot`, and the current behavior of `Snapshot::matches`/`Snapshot::matches_fully` is moved to a default `Comparator` implementation.

To compare PNGs, a new macro that is structurally identical to `assert_binary_snapshot` is introduced. Its chief difference is that it passes a `Comparator` that decodes image data and does a comparison on the basis of the decoded data.

This design is not perfect. In particular:

1. The coupling between snapshot contents and `Comparator` is loose. It might arguably be better for snapshot metadata to record how the comparison should be done. But I imagine that this would necessitate some sort of internal registry of `Comparator`s (or an analogous comparison function type), with `Comparator` selection done on the basis of a field in snapshot metadata. It seems simpler to keep `Snapshot`s concerned with the data (binary or otherwise) that is kept on disk and for *behavior* about how to handle `Snapshot`s to be specified at an assertion macro call site.
[...]

PR #861, opened 2026-01-26
Add Comparator trait and assert_png_snapshot macro

dstu:
Per #860, this is a prospective implementation of `Comparator` and `assert_png_snapshot`.

This PR factors the behavior of comparing two `Snapshot`s out into a `Comparator` trait. Existing comparison behavior is preserved in a default `Comparator`, and a new `Comparator` (and the `assert_png_snapshot` macro) are provided for comparing the image data of two PNGs. This is motivated by the need to decompress the PNGs before comparing the data directly. (For more discussion, see #860.)


max-sixty:
does each `Comparator` need to be in insta? I think that makes this much less viable to add to insta.

dstu:
This prospective implementation is hidden behind a feature in the main insta crate for convenience. It is straightforward to implement PNG comparison support in a different crate, given the other changes in this PR.

This would entail extending insta's public API to include the `Comparator` trait, however. And functionality provided by macros in `insta::_macro_support` (which are called by `assert_png_snapshot` in this PR) would need to be made more of an official public interface or vendored.

How do you feel about extending the public interface like this? Given the comments that currently exist about `assert_binary_snapshot` being experimental, I didn't want to make sweeping changes to insta's public API out of the blue.

max-sixty:
I'm wary about introducing additional features that aren't broadly helpful / others can't build on. I understand this proposal is behind a cargo feature, but it still adds some weight and maintenance burden to the library

> This would entail extending insta's public API to include the `Comparator` trait, however. And functionality provided by macros in `insta::_macro_support` (which are called by `assert_png_snapshot` in this PR) would need to be made more of an official public interface or vendored.

my question (and sorry not to have thought more about this myself, am trying to respond quickly even if not completely) is whether we could have a setting where a closure could be provided which would do the comparison. would we need a trait at all? if we did need a trait, for `matches` & `matches_fully`, would that be viable?

...then it's more likely that this would be generally helpful / would make insta a better library

(I'm not sure it's simple, would need to investigate. But without something general, I would be -1 on merging this, though would weigh @mitsuhiko's views higher than mine... )

dstu:
(following up... thanks for the prompt responses earlier, and I hope the delay hasn't been too long!)

> my question (and sorry not to have thought more about this myself, am trying to respond quickly even if not completely) is whether we could have a setting where a closure could be provided which would do the comparison. would we need a trait at all? if we did need a trait, for matches & matches_fully, would that be viable?

The short answer: I think a trait with a single method, as proposed, here, makes the most sense. Some other trait or closure would be sufficient. And some of the support macros that are currently called by `assert_binary_snapshot` should become part of a more official public interface.

Longer answer:

A plain old `&dyn Fn` would suffice, but this PR proposes a `Comparator` trait. My experience with interfaces like this one has led me to prefer a proper trait (a named type that can be extended, have associated doc comments, have old methods deprecated, etc.) over a simple closure. But the form of this interface is ultimately a decision for the insta maintainers.

It also proposes a single method (`Comparator::matches`) instead of two distinct methods (`matches` and `matches_fully`) because it seems like the distinction between `matches` and `matches_fully` is not something that should be baked into a generic interface. It is a distinction that matters for certain types of data but not all. The current behavior (use `matches_fully` when comparing two text snapshots and `require_full_match` is set) can be thought of as a runtime behavior of text-snapshot-specific snapshots. `Comparator::matches` takes a configuration object so that callers can control match behavior.

Going forward:

PNG comparison doesn't have to be in `insta` proper. If a suitable interface becomes available, I'd be content to maintain an `insta-png` crate (or something with a similarly appropriate name).

I'd be happy to provide a different PR with prospective interface changes. Would you like to see that?

Some more design discussion might be warranted, though. For example: in this PR, `Comparator::matches` just takes `Snapshot` parameters. But, by design, some `Comparator`s will likely be meant to work with some kinds of `Snapshot`s and not others. (In this PR, it is a runtime error to pass text `Snapshot`s to the PNG-specific `Comparator`.) It would be better for this class of error to be ruled out with static type checking. This might entail a more invasive change to insta, though - `Snapshot` kind is currently "erased" when `SnapshotValue`s are passed to `runtime::assert_snapshot`. Addressing this might lead to a general improvement in insta, but I didn't want to embark on a larger refactor without better motivation first.

max-sixty:
thanks for the message

don't want to give any false promises or hope, but I would be tentatively +0.5 if we could get a generic `Comparator` into insta's settings. I haven't thought through enough of whether there would be difficulties to be confident, though.

that would require splitting out the `png` feature so we could just see the implementation of the generic `Comparator`, allow it to be supplied in the settings' `bind`, etc.

one small note — I agree the `matches` vs `matches_fully` is somewhat legacy. But also we could handle it with a default implementation in the trait where `matches_fully` calls `matches`. I do think we should try and make small, non-intrusive changes, as you mention at the end of your comment; any refactors etc we can do separately.

dstu:
Ok, then I suppose this PR can be closed whenever it is no longer useful to continue discussion here. I'll close it once we have a thread going somewhere else. (You're free to close it once you're done here, too, of course.)

How much would it make sense to continue discussing design before sharing code? If so, where? A new feature issue would probably make more sense than hashing out details of a prospective design here.

I could also just factor interface changes out of this PR (which is already has a distinct commit for an interface change) and open a new PR based on those changes, with the aim of doing design review there.

max-sixty:
as you wish re using this vs opening another PR

I think it would be most useful to see proposed code, to work out where the problems are likely to be... though if the problems come sooner than expected, feel free to raise the issues sooner.

thanks!

dstu:
I have opened #872, so I'll close here.

PR #872, opened 2026-02-03
Add Comparator trait

dstu:
Per discussion in #861, here is a prospective implementation of a `Comparator` trait.

There may be some unsettled design questions. (Maybe just whether `ToolConfig` should be exposed so that we don't need to add the crufty `matches`/`matches_fully` distinction in `Comparator`.) As-is, this passes `make test`.

### DIFF
--- a/insta/tests/test_comparator.rs
+++ b/insta/tests/test_comparator.rs
@@ -1,11 +1,8 @@
-//! Provides end-to-end tests of custom [`Comparator`] implementations.
-//!
-//! If you're interested in writing macros with custom snapshot comparison
-//! behavior, consult these examples.
+//! Provides end-to-end tests of an example custom [`Comparator`]
+//! implementation. If you're interested in writing macros with custom snapshot
+//! comparison behavior, consult these examples.
 
 use insta::{assert_snapshot, with_settings, Comparator, Snapshot};
-
-// --- Custom comparator: whitespace insensitive ---
 
 /// A comparator that ignores whitespace differences.
 struct WhitespaceInsensitiveComparator;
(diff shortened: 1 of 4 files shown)


insta/src/runtime.rs
```diff
@@ -908,17 +908,16 @@ pub fn assert_snapshot(
...
-            }
-        })
-        .unwrap_or(false);
+    let pass = Settings::with(|settings| {
+        ctx.old_snapshot
+            .as_ref()
+            .map(|x| {
+                settings
+                    .comparator()
+                    .matches(&ctx.tool_config, x, &new_snapshot)
```
max-sixty: do we lose `matches_fully`?

max-sixty:
@dstu if you want to merge something tiny into the repo, even a linebreak, then you can run CI automatically and iterate to get tests passing.

Otherwise I need to click each time (and don't have permissions to change how this works). Otherwise you could run it on your own fork first.

dstu:
> We would need lots of functional tests added to see how this works in practice. Agents are fairly good at setting those up if you write the spec & review the outputs...

I have provided tests that include using a custom `Comparator`. Do you think something further is warranted? (Maybe a more complex `Comparator`?)

> if you want to merge something tiny into the repo, even a linebreak, then you can run CI automatically and iterate to get tests passing.
> 
> Otherwise I need to click each time (and don't have permissions to change how this works). Otherwise you could run it on your own fork first.

I'd be happy to run CI locally to fix issues so that you don't have to keep re-running it, but I'm not sure what you mean by running CI in my own fork. I'll run the Makefile-based commands from the GitHub workflows, though.

dstu:
RE the failing tests: the rustfmt check is easy to fix, but the clippy check has to do with the use of `Arc` when `Comparator` is not `Send` and `Sync`. I think that `Arc` can be replaced entirely with `Rc` in `settings.rs`, which should address the problem. Or the shared-settings-that-are-cloned-on-write model could be dropped entirely, in favor of just putting fields into `Settings`.

max-sixty:
for the moment, can we either ignore the clippy error so we can get the rest of the tests run, or make a separate PR making that change?

(ty!)

max-sixty:
I made some changes, add a bunch of functional tests; feel free to adjust

insta/tests/test_comparator.rs
```diff
@@ -1,65 +1,76 @@
...
+    // The value has single spaces, reference has multiple - custom comparator should match
+    let value = "hello world";
+    with_settings!({comparator => WhitespaceInsensitiveComparator}, {
+        assert_snapshot!(value, @"hello    world");
     });
 }
 
-#[test]
-#[should_panic(expected = "snapshot assertion for 'custom_comparator_fails' failed")]
-fn custom_comparator_fails() {
```
dstu: Here and elsewhere, you have removed tests which assert that snapshot tests fail when a custom `Comparator` fails to match them. This reduces test coverage. Do you not want that branch tested in this PR? (I realize that this may in part be due to `cargo insta test` marking tests as failed even when they are annotated with `should_panic`. Would it help if I open an issue for that?)

dstu:
> I made some changes, add a bunch of functional tests; feel free to adjust

Thanks. I think we're about ready to submit, but I'd like to make another round of changes to clean things up.

I'd like to ensure that there isn't too much back-and-forth ("I think it should look like this"/"But I think it should look like that"). I'm accustomed to leaving comments on commits to discuss things before making changes, but I can also just make a new commit with my preferred changes or edit commits you have made. What is your preference here?

max-sixty:
> I'd like to ensure that there isn't too much back-and-forth ("I think it should look like this"/"But I think it should look like that"). I'm accustomed to leaving comments on commits to discuss things before making changes, but I can also just make a new commit with my preferred changes or edit commits you have made. What is your preference here?

totally as you wish. you're ofc welcome to roll back my changes! sorry if it felt like I was overwriting your work...

dstu:
I'm torn because the commits from Claude make things seem simpler but worse for code health. As maintainer, if you are okay with things as-is, then let's just merge. I'll leave comments where I hope they will be helpful.

max-sixty:
ok, excellent feedback, and sorry for the scale of changes. I'm rolling some back.

I am keen to keep the `matches` / `matches_fully`; maybe we change that in the future but it does work well at the moment. and am keen to add functional tests, otherwise we get narrow bugs down the line, and "testing the tester" in unit tests inevitably leaves holes

everything else I have no strong view on, let me roll some of it back and your call on whether there's anything remaining that isn't good

thanks again for the code + the feedback

dstu:
+1 to keeping `matches`/`matches_fully`, at this point. Happy to review further and keep things moving!

max-sixty:
check those out @dstu — happy to roll back more if you think necessary...

dstu:
This looks good to me!

PR #873, opened 2026-02-04
Don't use `Arc` in `Settings` unnecessarily

dstu:
Per discussion in #871, it seems better for `Settings` to have an `Rc` internally instead of an `Arc`. This PR makes that change.

This passes `make test check-msrv lint check-minver format-check && cargo run -p cargo-insta -- test`.

Question for review: it would be simpler, and arguably better, if `Settings` just had fields instead of an inner `Rc`. This should leave the `Settings` interface as it is, since it uses simple interior mutation (fields would continue to be private with public accessors/mutators).

As-is, `Settings` has clone-on-write semantics, extra-cheap cloning (bumping a refcount copying fields), and extremely modest memory savings because newly constructed `Settings` objects share a set of immutable fields. But I'm not sure that the complexity imposed by having `Settings` wrap a pointer to an inner `ActualSettings` is justified. Is there a reason for using this pattern?

### DIFF
--- a/insta/src/settings.rs
+++ b/insta/src/settings.rs
@@ -1,4 +1,3 @@
-use once_cell::sync::Lazy;
 #[cfg(feature = "serde")]
 use serde::{de::value::Error as ValueError, Serialize};
 use std::cell::RefCell;
@@ -6,7 +5,7 @@ use std::future::Future;
 use std::mem;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
-use std::sync::Arc;
+use std::rc::Rc;
 use std::task::{Context, Poll};
 
 use crate::content::Content;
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws:
FYI, this is a breaking change (per cargo semver-checks)

```
    Building insta v1.47.0 (current)
       Built [   0.636s] (current)
     Parsing insta v1.47.0 (current)
      Parsed [   0.009s] (current)
    Building insta v1.46.0 (baseline)
       Built [   4.836s] (baseline)
     Parsing insta v1.46.0 (baseline)
      Parsed [   0.008s] (baseline)
    Checking insta v1.46.0 -> v1.47.0 (minor change)

--- failure auto_trait_impl_removed: auto trait no longer implemented ---

Description:
A public type has stopped implementing one or more auto traits. This can break downstream code that depends on the traits being implemented.
        ref: https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.47.0/src/lints/auto_trait_impl_removed.ron

Failed in:
     Checked [   0.006s] 196 checks: 194 pass, 2 fail, 0 warn, 56 skip
  type Redactions is no longer Send, in /tmp/insta-1.47/insta/src/settings.rs:26
  type Redactions is no longer Sync, in /tmp/insta-1.47/insta/src/settings.rs:26
  type SettingsBindDropGuard is no longer Sync, in /tmp/insta-1.47/insta/src/settings.rs:666
  type Settings is no longer Send, in /tmp/insta-1.47/insta/src/settings.rs:190
  type Settings is no longer Sync, in /tmp/insta-1.47/insta/src/settings.rs:190
  type Redaction is no longer Send, in /tmp/insta-1.47/insta/src/redaction.rs:53
  type Redaction is no longer Sync, in /tmp/insta-1.47/insta/src/redaction.rs:53

--- failure inherent_method_missing: pub method removed or renamed ---

Description:
A publicly-visible method or associated fn is no longer available under its prior name. It may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.47.0/src/lints/inherent_method_missing.ron

Failed in:
  Snapshot::matches, previously in file /home/jkastner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.46.0/src/snapshot.rs:516
  Snapshot::matches_fully, previously in file /home/jkastner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.46.0/src/snapshot.rs:525

     Summary semver requires new major version: 2 major and 0 minor checks failed
    Finished [   6.538s] insta
```

(discussion appears to be in #872)

max-sixty:
is anyone affected by this? or it's just theoretical?

john-h-kastner-aws:
There's a small break in our tests (https://github.com/cedar-policy/cedar-spec/issues/917). We can work around it, but it's enough to block an automatic update.

max-sixty:
ok. if you would like a reversion, happy to make it. sorry for the trouble

max-sixty:
@john-h-kastner-aws I have reverted. Sorry again for the trouble

max-sixty:
@dstu this makes the comparator slightly heavier. But the backward-compat is important, and seems like folks were relying on it. Hope that's OK.

dstu:
Thanks for the heads-up. This doesn't appear to affect the insta-image crate, so things look good from that end.

PR #883, opened 2026-03-08
the test_glob test depends on both glob and json

alexanderkjall:
This makes the command `cargo test --no-default-features --features glob` work


max-sixty:
thank you @alexanderkjall !

PR #885, opened 2026-03-14
Upgrade `console` to 0.16, bump MSRV to 1.66

max-sixty:
Upgrades `console` from 0.15.4 to 0.16 in both `insta` and `cargo-insta`, avoiding duplicate compilations for downstream projects that already depend on console 0.16 (e.g. via indicatif).

MSRV bumps from 1.64/1.65 to 1.66 — the minimum required by console 0.16.0. Console is pinned to 0.16.0 in the lockfile since 0.16.1+ requires Rust 1.71. The `std` feature is explicitly enabled for console in insta, as 0.16 gates `style`/`Term` behind it.

Supersedes #790

> _This was written by Claude Code on behalf of @max-sixty_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,7 +73,7 @@ version = "1.46.3"
 dependencies = [
  "cargo_metadata",
  "clap",
- "console",
+ "console 0.16.0",
  "ignore",
  "insta",
  "itertools",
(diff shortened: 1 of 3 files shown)


max-sixty:
this is consistent with a norm of MSRV of a version released three years ago. it's important for the project to be conservative on this. 

though probably this will be less salient when the MSRV-aware resolver is available for the older rust versions.

let me know any thoughts or feedback — if a _more_ conservative policy would be helpful, happy to update

Issue #868, opened 2026-01-31
List all snapshots?


tisonkun:
Recently, I met a requirement to list all snapshots and process them with another tool.

Some/Most of my snapshots are inlined. So I can't easily list snapshot files. Also, I'd like to process the snapshot content, rather than the metadata (filename, expr, etc.).

Is there a way to list all snapshots under a .. local crate?

max-sixty:
I have been coincidentally thinking about this.

I just pushed #871

_but_ we can't get the inline snapshots without compiling & running code; so it only lists file snapshots

we could consider adding a `--include-inline` flag, though it would be a discontinuous change in the work that `cargo insta show` does, from a file-listing to a full compilation...

max-sixty:
@tisonkun if you have the patience: can you try https://github.com/mitsuhiko/insta/pull/871 and see if it does what you'd like?

tisonkun:
@max-sixty Is there an easy way I can test out an unrelease binary? I forget how to do it :P

tisonkun:
@max-sixty Ah. I remember to install with `cargo install --path <path>`.

It lists all snapshot files but no inline ones. Depend on your design, it may or may not desired.

Sorry I can't remember why I need to list all snapshots now. But from my previous comment, I believe I want something like:

```json
[
  {"name": ..., "content" : "<snapshot content without metadata>" ... },
  {"name": ..., "content" : "<snapshot content without metadata>" ... },
]
```

or just jsonlines or any other reasonable format.

PR #897, opened 2026-04-15
Support clap long version

tisonkun:
cc @max-sixty 

```
$ cargo run --bin cargo-insta -- --version
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/cargo-insta --version`
cargo-insta 1.47.2
branch:clap-long-version
commit_hash:0843c8c
build_time:2026-04-16 06:42:35 +08:00
build_env:rustc 1.96.0-nightly (38c0de8dc 2026-02-28),nightly-aarch64-apple-darwin
```

When debugging #868, I noticed that a long version info with commit_hash helps or else I got only `cargo-insta 1.46.3` where I'm unsure if the used tool is from a local build.

Note that the `branch` and `commit` fields are empty if installed via `cargo install cargo-insta` because the published crate has no .git info. But that should be fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


max-sixty:
thanks @tisonkun !

do we need a whole new dependency for this? we would need to resolve the MSRV issue...

tisonkun:
@max-sixty I was also noticing this after seeing the CI failure. 

I'm not quite sure about MSRV. Because `cargo-insta` is a binary rather than a software library that needs to be depended upon. Consequently, I'm unclear on this project's MSRV policy. 

I decided to leave it as is for now, but we do need to bump the MSRV to accept this PR.

tisonkun:
Seems not `shadow-rs`'s issue but `time`'s issue that used in `build-dependencies`.

Let me see if I can relax the MSRV requirement.

tisonkun:
I'm going to see if `shadow-rs` can depend on `jiff` but it also requires MSRV >= 1.70.0, while `cargo-insta` has MSRV=1.66.0.

Pin `time` can't work because `shadow-rs` requires `time >= 0.3.47` while only `time <= 0.3.23` fits `MSRV <= 1.66.0`.

Ref - https://github.com/baoyachi/shadow-rs/pull/259

tisonkun:
https://github.com/scopedb/scopeql/pull/30

I have a solution to avoid extra dep. Let me see if I can bring it here later.

tisonkun:
Using gix would require MSRV=1.82. I think we can fall back to Git command and try to populate the info with best effort.

Issue #15, opened 2019-02-08
Presence of a build script causes "error: No such file or directory" after "cargo insta review"


jonas-schievink:
Repro on the `file-not-found` branch here: https://github.com/jonas-schievink/instabug/tree/file-not-found

Running `cargo test` will fail as expected, and touch the `.snap.new` file. Then, running `cargo insta review` will correctly show the diff between the versions. However, accepting the new snapshot just shows the same snapshot *again*, and accepting it again will fail with `error: No such file or directory (os error 2)`.

For some reason this is caused by `build.rs`. Removing it fixes the issue.

PR #904, opened 2026-05-13
Fix `--snapshot` filter on Windows; allow partial paths

### LINKED ISSUE #902: Bad windows path in `--snapshot` filter
### What happened?

The `--snapshot` option in `review`/`accept`/`reject` does not support relative path or partial name, at least on Windows. I did not test on Linux, though.

It is annoying that I have to expose the absolute path to it, and the Windows path format is quite noisy.

This problem only arises when LLMs attempt to alter the snapshots according to instructions from `cargo insta pending-snapshots`. But they often fail because of the naturally used relative paths.
I think it will not be a problem for humans, as normal users probably would not use this method to process snapshots one by one...

### Reproduction steps

```
❯ cargo insta pending-snapshots
Pending snapshots:
  \\?\C:\path\to\shot.snap

To review a snapshot: cargo insta review --snapshot '\\?\C:\path\to\shot.snap'
To accept a snapshot: cargo insta accept --snapshot '\\?\C:\path\to\shot.snap'
To reject a snapshot: cargo insta reject --snapshot '\\?\C:\path\to\shot.snap'

To review all interactively: cargo insta review
To accept all: cargo insta accept
To reject all: cargo insta reject

❯ cargo insta accept --snapshot shot.snap
insta review finished
skipped:
  path/to/tests.rs (shot.snap)
```

### rustc Version

_No response_

### insta Version

_No response_

### cargo insta --version output

1.47.2

### What did you expect?

Support partial name as the snapshot filter.

max-sixty:
On Windows, `cargo insta pending-snapshots` printed snapshot paths with the `\\?\` verbatim prefix (e.g. `\\?\C:\path\to\shot.snap`), and `--snapshot <path>` required an exact match against that hard-to-produce string — frustrating for users and especially for LLM tooling that naturally produces relative paths.

Root cause: `find_snapshot_roots` canonicalised search roots, and on Windows `canonicalize()` returns verbatim paths. Those flowed into `target_path` on each `SnapshotContainer`, and `format_snapshot_key`'s `strip_prefix(workspace_root)` then failed (because `workspace_root` is *not* verbatim), so the full `\\?\C:\…` string was emitted.

Fix:

- `find_snapshot_roots` now keeps the GH-15 dedup but compares *canonicalised* forms while returning the *original* (non-verbatim) paths. This also lets `strip_prefix_with_fallback` hit its symlink-preserving direct-strip path for Bazel/`INSTA_PENDING_DIR` mode rather than falling through to canonicalise-both.
- `format_snapshot_key` reuses the existing `insta::utils::path_to_storage` helper so cargo-insta and insta agree on what a stored path looks like (workspace-relative, `/` separators on every platform).
- The `--snapshot` filter is now lenient: a value matches when it is a trailing path-component suffix of the snapshot's target file, so a bare `foo.snap` (or `tests/snapshots/foo.snap`, or the full key) all work. Implemented via `Path::ends_with`, which already treats `/` and `\` as equivalent in the needle on Windows. Inline snapshots may carry an optional `:line` suffix to disambiguate.

Tests: unit tests for `format_snapshot_key` and `snapshot_matches_filter` plus two functional tests (bare-name `accept --snapshot` accepts only the matching snapshot; `pending-snapshots` keys are workspace-relative and `/`-separated).

Closes #902. Thanks to @QuadnucYard for reporting — would you mind giving this a try on Windows once it lands? In particular: `cargo insta pending-snapshots` should now print clean relative paths, and `cargo insta accept --snapshot my_test.snap` (or any trailing-path-suffix) should work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -412,6 +412,7 @@ dependencies = [
  "serde",
  "similar",
  "similar-asserts",
+ "strip-ansi-escapes",
  "tempfile",
  "toml_edit",
  "toml_writer",
@@ -805,6 +806,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "strip-ansi-escapes"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2a8f8038e7e7969abb3f1b7c2a811225e9296da208539e0f79c5251d6cac0025"
+dependencies = [
+ "vte",
+]
+
+[[package]]
 name = "strsim"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -961,6 +971,15 @@ name = "version_check"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "49874b5167b65d7193b8aba1567f5c7d93d001cafc34600cee003eda787e483f"
+
+[[package]]
+name = "vte"
+version = "0.14.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "231fdcd7ef3037e8330d8e17e61011a2c244126acc0a982f4040ac3f9f0bc077"
+dependencies = [
+ "memchr",
+]
 
 [[package]]
 name = "walkdir"
--- a/cargo-insta/src/cargo.rs
+++ b/cargo-insta/src/cargo.rs
@@ -1,7 +1,6 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 pub(crate) use cargo_metadata::Package;
-use itertools::Itertools;
 
 /// Find snapshot roots within a package
 // We need this because paths are not always conventional — for example cargo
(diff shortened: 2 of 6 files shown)


QuadnucYard:
Yeah, it works now!

But the message still seems a little misleading:
```
❯ cargo run -- accept --snapshot test_basic
insta review finished
skipped:
  insta/tests/test_basic.rs (unnamed_debug_vector.snap)

❯ cargo run -- accept --snapshot unnamed_debug_vector.snap
insta review finished
skipped:
  insta/tests/test_basic.rs (unnamed_debug_vector.snap)

❯ cargo run -- accept --snapshot test_basic__unnamed_debug_vector.snap
insta review finished
accepted:
  insta/tests/test_basic.rs (unnamed_debug_vector.snap)
```

- The message does not imply that the filter does NOT hit any snapshot.
- It does not give the right name in the parens that can be used as the filter for that snapshot.
- It would be better to suggest what filter to use when it misses.
  - Advanced feature: use the shortest suffix (path segments) that can distinguish this snapshot as a suggestion, e.g., when multiple snapshots share the same stem.

This might require another PR to improve, though.

PR #907, opened 2026-05-17
fix(redactions): apply sorted_redaction after other redactions

### LINKED ISSUE #683: Redaction order seems to matter when using `sorted_redaction`
### What happened?

When using the `sorted_redaction` it seems to be important to place the sorting redactions at the end which I did not expect to matter. Also using `[]` as the selector instead of `.` also seems to fix the issue in the example below but does not seem to work if the redacted key value pair is nested somewhere deeper.

### Reproduction steps

```rs
// insta = { version = "1.41.1", features = ["json", "redactions"] }
// rand = "*"
// serde_json = "*"
#[test]
fn insta_sorting_bug() {
    use insta::assert_json_snapshot;
    use rand::random;
    use serde_json::json;

    let json = json!([
        {
            "a": random::<bool>(),
            "z": 1,
        },
        {
            "a": random::<bool>(),
            "z": 2,
        }
    ]);
    // Inconsistent
    assert_json_snapshot!(json, {
        "." => insta::sorted_redaction(),
        "[].a" =>  "[redacted]",
    });
    // Consistent for this case but not for more deeply nested redacted values
    // assert_json_snapshot!(json, {
    //     "[].a" =>  "[redacted]",
    //     "." => insta::sorted_redaction(),
    // });
    // Consistent
    // assert_json_snapshot!(json, {
    //     "[]" => insta::sorted_redaction(),
    //     "[].a" =>  "[redacted]",
    // });
}
```

### Insta Version

1.41.1

### rustc Version

1.81

### What did you expect?

Redaction order to not be important or its importance mentioned in the docs of `sorted_redaction`.

wyf027:
## Summary

Fixes #683.

`sorted_redaction()` is now applied in a second pass after all other redactions, so snapshot output no longer depends on the order entries appear in the redaction map.

## Changes

- Add `Redaction::Sort` variant used by `sorted_redaction()`
- `Redactions::apply_to_content` runs non-sort redactions first, then sort redactions
- Document that sort redaction order in the map is irrelevant

## Test plan

- [x] `cargo test -p insta --lib --features redactions,json sort_redaction_order`
- [x] `cargo test -p insta --test test_redaction --features redactions,json,yaml`
- [ ] CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -412,6 +412,7 @@ dependencies = [
  "serde",
  "similar",
  "similar-asserts",
+ "strip-ansi-escapes",
  "tempfile",
  "toml_edit",
  "toml_writer",
@@ -805,6 +806,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "strip-ansi-escapes"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2a8f8038e7e7969abb3f1b7c2a811225e9296da208539e0f79c5251d6cac0025"
+dependencies = [
+ "vte",
+]
+
+[[package]]
 name = "strsim"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


max-sixty:
thanks! I think this makes sense.

are there any backward compat concerns? I don't think there should be...

PR #909, opened 2026-05-17
feat(cargo-insta): add `cargo insta diff`

### LINKED ISSUE #886: Command to show the diff of a single pending snapshot
Sometimes I want to look at a specific diff without having to tap my way through `cargo insta review`'s interactive review to find the right one.

With `cargo insta pending-snapshots` I can find all the snapshots that are pending, but there is currently AFAIK no way to easily see the diff for one of the pending snapshots.

I've been using this command (with [wdiff](https://www.gnu.org/software/wdiff/)):

```sh
wdiff <(cargo insta show my-crate/tests/snapshots/test__test212.snap) <(cargo insta show my-crate/tests/snapshots/test__test212.snap.new)
```
but it's a lot to type and it doesn't have colors and it shows too much of the surrounding text.

wyf027:
## Summary

Adds `cargo insta diff` to show colored pending snapshot diffs without interactive review.

- `cargo insta diff` — all pending snapshots
- `cargo insta diff --snapshot <key>` — one or more filtered snapshots (same keys as `pending-snapshots`)

`pending-snapshots` output now suggests `cargo insta diff --snapshot` first.

Fixes #886

## Test plan

- [x] `cargo test -p cargo-insta` (103 tests passed)

### DIFF
--- a/cargo-insta/src/cli.rs
+++ b/cargo-insta/src/cli.rs
@@ -665,11 +665,8 @@ fn print_snapshot_diff(
     println!("  Package: {}@{}", package.name.as_str(), &package.version);
     println!();
 
-    let mut printer = SnapshotPrinter::new(
-        workspace_root,
-        snapshot_ref.old.as_ref(),
-        &snapshot_ref.new,
-    );
+    let mut printer =
+        SnapshotPrinter::new(workspace_root, snapshot_ref.old.as_ref(), &snapshot_ref.new);
     printer.set_snapshot_file(snapshot_file);
     printer.set_line(snapshot_ref.line);
     printer.set_show_info(true);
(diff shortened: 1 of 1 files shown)


max-sixty:
WDYT about having `cargo insta show --diff`? That would reduce any API sprawl? (I'm not confident it's a good idea though)

wyf027:
Addressed the CI failures in commit 0c4eddd:

- Applied the rustfmt change reported by the rustfmt job for cargo-insta/src/cli.rs.
- Collapsed the nested shown == 0 / !quiet branch reported by clippy.

I also saw the API-sprawl comment about cargo insta show --diff. I kept this as cargo insta diff for now because the command is a non-interactive diff-only workflow rather than an extension of show output, but I can switch it to show --diff if that better fits the CLI direction.

wyf027:
Resolved the merge conflict with upstream master in commit 070b6e1. I kept the upstream snapshot filter helper and wired `cargo insta diff --snapshot ...` through the same matcher, so the diff command follows the current filter semantics.

Local verification:
- cargo fmt --check
- cargo test -p cargo-insta snapshot_filter --test functional
- cargo test -p cargo-insta test_snapshot_matches_filter
- cargo test -p cargo-insta test_format_snapshot_key

PR #914, opened 2026-05-30
fix: tolerate a missing binary snapshot data file

### LINKED ISSUE #911: Binary snapshot in .gitignore not seen even with --include-ignored
### What happened?

Hello, I wanted to share a use case. I've included the binary snapshots in `.gitignore` by file extension because they're too big and I only use them as a way to produce final example files (`.xlsx` in this case) to open them in another program.

The snapshot metadata files, on the other hand, are not in `.gitignore` because there is generally no issue with keeping unused snapshot files around.

However when I run `cargo insta accept`, even with `--include-ignored`, the CLI fails with nonspecific `error: No such file or directory (os error 2)`.

My best guess is that this is because it sees the metadata file even after filtering out ignored files, and doesn't try to look for the binary file among the ignored files.

### Reproduction steps

_No response_

### rustc Version

1.94.1

### insta Version

1.47.2

### cargo insta --version output

cargo-insta 1.43.1

### What did you expect?

Ideally I think the optimal behavior would be for the CLI to ignore the `.gitignore` filters when looking for the binary files, as the literal path is already included in the metadata file.

If it is necessary to prevent the ignored file from being seen in this way, it would be good for the error to be explicit about the file existing but being ignored.

max-sixty:
`cargo insta accept` (and `cargo insta test --accept`) fails with a nonspecific `No such file or directory (os error 2)` and refuses to accept whenever a tracked `.snap` binary snapshot's data file is absent, even with `--include-ignored`. That absence is the normal state after checking out a project that gitignores its binary data files: the `.snap` metadata is committed but the data files are not, so they're missing on any fresh clone, on CI, or after `git clean`. This is the setup reported in #911.

The cause isn't the ignore walker, which is why `--include-ignored` doesn't help. When a snapshot is loaded, `Snapshot::from_file` eagerly `fs::read`s the binary sidecar, and a missing sidecar makes that `?` abort the whole load. But `cargo-insta` only needs the old snapshot's metadata here (it never reads the old binary's bytes, only the extension, to clean up a stale file), and insta's own runtime already tolerates a missing old snapshot during `cargo insta test`. So the abort was an inconsistency, and one you couldn't recover from with the obvious `accept`.

This makes an absent payload a first-class state. `SnapshotContents::Binary` now holds `Option<Rc<Vec<u8>>>`; `from_file` maps a `NotFound` data file to `None` instead of erroring (a present-but-unreadable file still errors). A freshly produced snapshot always carries `Some`. This fixes both the `accept` crash and the misleading `Failed to parse snapshot file` warning the runtime printed in the same scenario.

Note: this changes the shape of the public `SnapshotContents::Binary` variant. No call site in either crate binds the payload directly (all match via `_`, `{ .. }`, or `unreachable!()`), so nothing else needed touching, but it's a public-type change worth a semver note.

Includes a functional test (`test_binary_accept_missing_old_binary`) that reproduces the failure (tracked `.snap`, absent data file, then `accept`) and passes with the fix.

Closes #911. Thanks to @Kinrany for the report.

> _This was written by Claude Code on behalf of max_

### DIFF
--- a/cargo-insta/tests/functional/binary.rs
+++ b/cargo-insta/tests/functional/binary.rs
@@ -1,3 +1,6 @@
+use std::fs;
+use std::process::Command;
+
 use insta::assert_snapshot;
 
 use crate::TestFiles;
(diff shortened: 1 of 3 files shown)


Kinrany:
Thank you!

PR #916, opened 2026-06-02
Improve `--snapshot` filter feedback in cargo-insta

max-sixty:
Follow-up to #904, addressing @QuadnucYard's feedback there.

Two related fixes to how `cargo insta accept/reject/review --snapshot` reports what it did:

The accepted/rejected/skipped lists now print each snapshot's canonical key (the same workspace-relative path `pending-snapshots` prints and `--snapshot` accepts) instead of the bare snapshot name. The bare name was never a usable filter, because the on-disk file is module-prefixed (`mod__name.snap`).

A `--snapshot` entry that matches nothing now produces a warning on stderr, rather than silently leaving every snapshot under `skipped:`. Exit status is unchanged, so a no-op accept still succeeds. The pending keys to use instead are already in the `skipped:` list on stdout, so the warning doesn't repeat them.

The functional tests snapshot the full command output, so the new format is visible in the diff:

```
stdout:
insta review finished
skipped:
  src/snapshots/snapshot_filter_no_match__greeting.snap

stderr:
warning: --snapshot 'greeting.snap' didn't match any pending snapshot
```

Thanks to @QuadnucYard for the suggestions in #904.

> _This was written by Claude Code on behalf of max_

### DIFF
--- a/cargo-insta/src/cli.rs
+++ b/cargo-insta/src/cli.rs
@@ -714,6 +714,10 @@ fn review_snapshots(
     let mut accepted = vec![];
     let mut rejected = vec![];
     let mut skipped = vec![];
+    // Track which `--snapshot` filter entries matched something, so we can warn
+    // about entries that hit nothing. The leftover `skipped` keys then double as
+    // the list of still-pending snapshots to suggest instead.
+    let mut matched_filters = vec![false; snapshot_filter.map_or(0, |f| f.len())];
     let mut num = 0;
     let mut show_info = true;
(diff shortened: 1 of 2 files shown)


max-sixty:
@QuadnucYard would you mind taking a look? max thought your suggestions on #904 were good, so this PR implements them: snapshots are now listed by the key `--snapshot` actually accepts, and a filter that matches nothing warns and shows the valid keys instead of silently skipping everything.

> _This was written by Claude Code on behalf of max_

PR #924, opened 2026-06-10
Allow `CI=true` to be overridden by an explicitly passed `--accept` CLI flag

AlexWaygood:
We have a CI job over at Ruff that automatically creates a PR updating some vendored files every two weeks. Updating the vendored files often means that snapshots need to be updated, and I'd like that to be done as part of the automated workflow using `--accept`. However, currently if the `CI` environment variable is set to a truthy value (which it always is in the context of GitHub Actions), that cannot be overridden using `--accept`.

That seems like an oversight in https://github.com/mitsuhiko/insta/pull/345: normally, CLI flags take precedence over environment variables. This PR fixes that oversight

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to insta and cargo-insta are documented here.
 
 ## Unreleased
 
+- Setting `CI=true` normally makes `cargo insta test` behave as though `--check`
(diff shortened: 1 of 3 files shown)


max-sixty:
thank you @AlexWaygood !

AlexWaygood:
Thanks for the quick release and the awesome tool @max-sixty! 🙌

PR #931, opened 2026-07-24
Introduce matches_snapshot

black-desk:
## Summary

Adds `matches_snapshot!`, the non-asserting counterpart of
`assert_snapshot!`. It accepts the same calling forms (named, auto-named,
inline `@"…"`) and applies the current bound `Settings` (filters,
redactions) to the value, comparing exactly as `assert_snapshot!` would —
but never writes a `.snap.new` file, never prints, and never panics. It
returns `Result<bool, Box<dyn Error>>`.

## Motivation

I'm developing a ratatui-based TUI application and want to test it the
`waitFor` way: drive some input, then poll the app's rendered state until
it reaches an expected snapshot, treating a timeout as failure (the same
"retry-until-it-passes" idea as [`waitFor`] in React Testing Library, or
Cypress's auto-retrying assertions).

In that loop, `assert_snapshot!` is actively harmful: **every**
intermediate non-matching poll writes a `.snap.new` to disk and panics, so
the natural `while !snapshot_matches { retry }` pattern instead requires
`catch_unwind` around the macro and litters the tree with pending
snapshots for each frame that didn't match yet.

`matches_snapshot!` closes that gap — a `while !matches_snapshot!(…)` loop
that reports the verdict with no side effects, leaving the single failure
(if the value never converges) to one final `assert_snapshot!` that *does*
write the `.snap.new` for review:

```rust,ignore
let deadline = Instant::now() + Duration::from_secs(5);
loop {
    drive_some_input(&mut app);
    if matches_snapshot!("my_snap", render(&app)).unwrap_or(false) {
        break;
    }
    if Instant::now() > deadline {
        // One real failure for review:
        assert_snapshot!("my_snap", render(&app));
    }
}
```

[`waitFor`]: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

## Implementation

Split into two commits:

1. **refactor** — Splits the internal `_assert_snapshot_base!` macro into
   `_snapshot_base!` (workhorse: normalizes the call form, returns the
   runtime `Result` without `.unwrap()`) and `_assert_snapshot_base!`
   (thin wrapper that unwraps, preserving panic-on-mismatch). Extracts
   three runtime helpers out of `assert_snapshot`: `build_text_contents`
   (ANSI stripping + filters), `build_snapshot_contents` (text/binary
   dispatch + extension validation, returns `Result`), and
   `snapshot_matches` (comparison). Pure refactoring; `assert_snapshot` is
   the only caller and its behavior is unchanged.

2. **feat** — Adds a `func` parameter to `_snapshot_base!` selecting the
   runtime function; `_assert_snapshot_base!` injects `func =
   assert_snapshot`, `matches_snapshot!` injects `func = matches_snapshot`,
  so the new macro inherits every calling form for free. The
  `matches_snapshot` runtime function mirrors `assert_snapshot` up to the
   comparison and then stops — skipping the pending-snapshot write,
   diff printing, and panic. (Like `assert_snapshot`, it still registers
   a file snapshot as referenced, so `cargo insta test --unreferenced`
   won't sweep a snapshot that is only ever checked, never asserted.)

   `SnapshotAssertionContext::prepare` gains an `asserting: bool` parameter
   that gates the two behaviors tied to writing a repaired snapshot:

   - the inline-duplicate guard (so inline `@"…"` is safe in a loop), and
   - the "failed to parse snapshot file" warning (an unparseable reference
     is treated the same as no reference — silently — since this path has
     nothing to regenerate).

   `assert_snapshot` passes `true` (unchanged behavior); `matches_snapshot`
   passes `false`.

## Tests

`insta/tests/test_matches_snapshot.rs` covers: matching value, mismatching
value (asserts no `.snap.new` is written), missing reference, filters
applied, the inline form, and inline form called in a loop (the polling
use case).

## Caveats

- **No binary counterpart.** `assert_binary_snapshot!` calls the runtime
  directly and does not go through `_snapshot_base!`, so there is no
  `matches_binary_snapshot!`. `matches_snapshot!` is Display/text-only,
  consistent with `assert_snapshot!`. Matching binary snapshots can be
  added later if there's demand; binary snapshots are still experimental.

## Commits
- **refactor: split snapshot base macro and extract runtime helpers**
- **feat: add side-effect-free matches_snapshot! macro**


max-sixty:
I empathize with the feature request, and it is a bit awkward that the main function panics rather than returns

but I do worry about proliferating macros — would we then need this for each variant of `assert`?

can I suggest trying:
- use the ability to avoid writing snapshots, same as `--check`
- wrap the existing macros in a `catch_unwind` as you suggest above

I would be more partial to a feature change like this if we can establish why that's preclusive, and we have some cases of the approach being helpful. Is that reasonable?

black-desk:
Thanks — I wasn't aware of --check/INSTA_UPDATE=no, much appreciated.

But the polling use case is inherently two-phase: during the loop you want it
silent (no file, no output), yet on timeout the final assertion does need to
write the .snap.new for review. So it needs to flip from no to new mid-run.
INSTA_UPDATE/INSTA_OUTPUT go through ToolConfig, which is cached on first use,
so they can't be toggled at runtime — a set_var after the first snapshot call is
ignored. Setting it globally up front means the timeout assertion inherits no
and produces no .snap.new at all, or setting it to new leaks into other tests
running in parallel.

On top of that there's the catch_unwind cost: each non-matching poll frame still
panics and prints the panic message to stderr (INSTA_OUTPUT=none only suppresses
insta's diff/summary, not the panic message itself), and the panic/unwind cycle
is measurable — in a quick benchmark of a 2000-iteration polling loop the
catch_unwind(assert_snapshot!) path ran about 2× slower than matches_snapshot!,
and that's before accounting for the stderr I/O.

PR #933, opened 2026-08-07
Add external text file snapshots

### LINKED ISSUE #475: Using insta for schema generation and updates

Turbo87:
`assert_file_snapshot!(path, value)` compares `Display` output with a plain text file at a caller-selected path. Relative paths resolve from the assertion source (like `include_str!(...)`). Accepted files contain only normalized text without Insta metadata.

External snapshots use the existing proposal and `cargo insta review` workflow. They remain outside unreferenced snapshot tracking, and dynamic targets receive stable pending identities.

The first two commits of this PR just prepare the codebase for the feature implementation in the latter two commits. This is best reviewed commit-by-commit. And don't be scared by the diff size, the majority of it is test cases 😅 

> [!NOTE]
> This was partially created by OpenAI Codex (GPT-5.6 Sol) and independently reviewed by myself and Anthropic Claude (Opus 5). My knowledge of the codebase is limited though, and the same goes for my ability to review this in depth. Feel free to tell me that this is all completely wrong and close the PR :)

I have another set of commits that builds an `assert_binary_file_snapshot!(...)` macro on top of this, if byte-for-byte equality is required, but I've left it our of this PR to keep the diff size at least somewhat manageable.

### Related

- Resolves #475 (at least partially)


max-sixty:
What's the motivation behind the change?

Turbo87:
the short version is that it somewhat mirrors https://vitest.dev/api/expect.html#tomatchfilesnapshot.

I often find myself in a situation where I want to generate derived code/data files from some source of truth. A recent example is that I had data in a JSON file and wanted to generate CSS and TypeScript code from that for the frontend project in the same repository. I ended up using a vitest test for that, but I've had similar situations in the past where keeping it in the Rust code would have made more sense, since I was generating the derived files with Rust libraries. 

The difference to the existing binary snapshot functionality in Insta is that I need these files in a specific location to be able to use them from there.

Integrating it into the build process might technically possible too in some cases, but is often more complex and doesn't allow reviewers to see the change in the output files. I worked around the issue so far by having an example bin write the files that a shared lib generates, and have a test that only checks for equality, but it would be great if a single test could just do both, like with e.g. Vitest.

max-sixty:
ok, the bar is quite high for adding things like this, so we would need quite compelling justification / we would need to think it more

PR #936, opened 2026-08-15
fix(redactions): enforce scalar total ordering in sorted_redaction

myukitty:
When sorting collections (`Seq`, `Map`, and `Struct` fields) in `sorted_redaction()`, floating-point values with NaNs and mixed-width integer keys (`U8`/`I8`) could violate transitivity and cause panics under `sort_by`.

This PR:
1. Sorts elements using `as_key()` with total ordering for scalar floats (`f64_total_cmp`).
2. Adds `Content::U8` and `Content::I8` variants to `as_key()` so same-signedness mixed-width integers and characters sort by numeric value.
3. Adds regression tests for mixed integer widths (unsigned and signed) and NaN floats.

Refs #587 (note: non-scalar `Key::Other` sorting fallback remains open in #587).


max-sixty:
looks good, thanks @myukitty

a couple of small suggestions from an LLM:

---

Approved the workflows — everything's green, including the MSRV job.

Two things before this goes in.

**The u8 test doesn't exercise the u8 change.** `test_sorted_redaction_u8_keys` passes on master too; I ran it with both source files reverted and only the tests applied. A homogeneous `HashMap<u8, _>` already sorted correctly, because `Key::Other` falls back to `Content`'s derived `PartialOrd`, which compares two `U8`s by value.

The `as_key` hunk is a real fix, it just needs a case that mixes integer widths to show it. This one does:

```rust
#[derive(serde::Serialize)]
#[serde(untagged)]
enum Mixed {
    N(u8),
    C(char),
    B(u64),
}

let v = vec![Mixed::N(200), Mixed::C('a'), Mixed::B(5)];
insta::assert_yaml_snapshot!(&v, { "." => insta::sorted_redaction() });
```

Master sorts that `200, 5, a`. With your change it's `5, a, 200`, which is the numeric order across the three widths (`'a'` is 97).

**This doesn't close #587.** The `Key::Other` arm still uses `partial_cmp(..).unwrap_or(Equal)`, so anything that isn't a scalar keeps the non-total order. A `Vec<Vec<f64>>` containing NaNs still panics, identically with and without this PR:

```rust
let nested: Vec<Vec<f64>> = (0..40)
    .map(|i| if i % 3 == 0 { vec![f64::NAN] } else { vec![(40 - i) as f64] })
    .collect();
insta::assert_yaml_snapshot!(&nested, { "." => insta::sorted_redaction() });
```

```
user-provided comparison function does not correctly implement a total order
```

(40 elements so the sort gets past the insertion-sort threshold, where that check actually fires.)

Fixing the flat case is worth having on its own, so could you change `Fixes #587` to `Refs #587` in the description? Otherwise merging closes an issue that's still live.

If you'd like to take the rest of it, one option: the comparator now appears four times, once in `sort_maps` and three times here. Lifting it into a single `pub(crate) fn content_cmp(&Content, &Content) -> Ordering` in `serialization.rs` would give the `Key::Other` case one place to become total, and all four call sites would pick it up. Optional, and I'm happy to merge the narrow fix without it.

> _This was written by Claude Code on behalf of max-sixty_

myukitty:
Thanks @max-sixty — agreed on both points!

1. Replaced the `u8`-only test with the mixed `u8`/`char`/`u64` case (`[200, 'a', 5]` -> `[5, 'a', 200]`) and added a signed mixed-width test (`[-1000, -50, 0, 12, 100]`) so the `as_key` integer changes are explicitly exercised.
2. Switched `Fixes #587` to `Refs #587` in the commit and description to keep #587 open for non-scalar `Key::Other` ordering.

Happy to open a follow-up for the unified `content_cmp` refactor if desired!

max-sixty (approved):
Thanks — the updated tests now exercise both unsigned and signed mixed-width ordering, and the description correctly leaves #587 open for the non-scalar fallback. The current head is clean, and the full Linux, macOS, Windows, MSRV, dependency, docs, lint, and self-test matrix passes.

> _This was written by Codex on behalf of max-sixty_

